PR #3, opened 2015-01-16
Review

chrisndodge:


chrisndodge:
@ormsbee mind taking a quick look at this repo? I'm getting ready to hand off this initial framework to our development team (who will be developing out notification triggers from the LMS into /notifications/api/publish.py).

In particular, I'd like a quick vetting of:

1) The store provider abstraction (goal: can we ultimately support a no-SQL backend)

2) The notification channel abstraction (goal: can we support multiple delivery channels, e.g. in-browser, triggered emails, SMS, mobile push notifications, etc.)

3) The data model abstraction as we can't use the Django ORM as the de-facto data model as we need to support different store providers

4) General implementation patterns

NOTE: This is not a complete implementation of the back-end, including the full definition of the abstract interfaces for Notification Channels and Notification Store Providers, as well as not a full schema definition for the Data Objects. If you'd rather wait for that full definition, I can do that, but it'll probably be another few days.

The imeplemtation is intended to be just enough for the developers to start the LMS notification triggers, so we can start some initial plumbing. Next up for me is building out the framework NotificationRendering, which is probably going to be the crux of the whole design. I'll tag you on when that is ready for an initial preview.

Also note, this is with our defined MVP in mind, so all that talk about AMPQ backend queueing is out of scope here, but it'd be good if you could vet that supporting such a backplane in the future would be feasible here via the Notification Channel abstraction.

FYI, @smagoun 

FYI, @afzaledx @muhhshoaib so that you can start to get a feel of the subsystem you'll be calling into. Today, I'll be writing up developer documentation in the README.md/Wiki area of this repo to give you enough information to let you all get started.

FYI, @mattdrayer @martynjames in case you are interested

notifications/store/sql/store_provider.py
```diff
@@ -0,0 +1,237 @@
...
+        _msg_type_cache_size = kwargs.get('MAX_MSG_TYPE_CACHE_SIZE', 1024)
+        self._msg_type_cache = pylru.lrucache(_msg_type_cache_size)
+
+    def _get_notification_by_id(self, msg_id, options=None):
+        """
+        Helper method to get Notification Message by id
+        """
+
+        # pylint/pep8 seem to complain if defaults are set to empty dicts
+        options = options if options else {}
```
ormsbee: It flags it because mutable default args are really unintuitive. If `options` was defaulted to be an empty dict in the method signature, it would hold onto that reference across multiple method calls. So if you ever mutated `options` inside the method, every successive call to that method would have the same mutated value for `options`. It makes sense when you think about how the definition process actually works, but it bites every Python programmer that I know of at least once.

ormsbee:
I'm not going to be able to do a deeper dive into this until later this week, but I'm concerned about the custom ORM-like layer that's being built. I totally get the desire for data objects that are not tied to a specific backend -- but can't those be plain old python objects? DRF will give you serializers that can handle the validation pieces.

chrisndodge:
@ormsbee I added some sample DRF serializers to see how it'd interoperate with this generic Data Object model. Seems to work with these two samples, but I'd need to write more unit tests to exercise it a bit harder.

edx_notifications/stores/store.py
```diff
@@ -0,0 +1,152 @@
...
+
+        ARGS:
+            - msg: an instance of NotificationMessage. If the 'id' field is
+                   set by the caller, then it is assumed to be an update
+                   operation
+
+        RETURNS: type NotificationMessage
+        """
+
+    def save_notification_user_map(self, user_map):
```
cpennington: This should probably be abstract

edx_notifications/stores/store.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        Create or Update the mapping of a user to a notification.
+        """
+
+    @abc.abstractmethod
+    def get_notification_type(self, name):
+        """
+        This returns a NotificationType object.
+        """
+
```
cpennington: It would be good to raise `NotImplementError` in these methods, so that if someone `super` chains up to them, they'll actually fail, rather than just returning `None`.

cpennington: I tend to follow the guidelines that abstract methods that change state should have empty base methods (no state change), and that abstract methods that return things should raise errors (no way to return an unknown empty thing from the base). And I try not to have methods that do both. =)

edx_notifications/api/consumer.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    Returns the number of notifications this user has
+
+    ARGS:
+        user_id: The id of the user to query
+        filters: A dict containing the following optional keys
+            - 'namespace': what namespace to search (defuault None)
+            - 'read': Whether to return read notifications (default True)
+            - 'unread': Whether to return unread notifications (default True)
+            - 'type_name': what msg_types to count
```
cpennington: Would it be better to just have these as optional arguments to this function? Or to have a `Filter` class that defines the set of valid filters?

edx_notifications/api/consumer.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    Returns a list of NotificationUserMap for the passed in user. The list will be
+    ordered by most recent first
+
+    ARGS:
+        - user_id: The id of the user
+        - filters: a dict containing
+            - msg_id: a particular notification id to retrieve
+                ( if this is used in conjuction with other filters, it may
+                  result in an empty set being returned )
```
cpennington: `msg_id` seems to always be incompatible with other filters, right? If you're querying for a specific message, it doesn't make sense to query anything else at the same time.

edx_notifications/api/consumer.py
```diff
@@ -0,0 +1,121 @@
...
+        # at least with SQL backends. However, for future no-SQL backends
+        # this is conceptually a possibility, so good to double check
+        err_msg = (
+            'There should be at most 1 NotificationUserMap items found for '
+            'msg_id {msg_id} and user_id {user_id}. {cnt} were found!'
+        ).format(msg_id=msg_id, user_id=user_id, cnt=len(notifications))
+
+        raise ItemIntegrityError(err_msg)
+
+    user_map = notifications[0]
```
cpennington: `user_map` is a confusing name for the results of `get_notifications_for_user`. Is there a better name for that concept? Perhaps `UserNotificationMessage` (in contrast to `NotificationMessage` for the raw message)?

edx_notifications/api/consumer.py
```diff
@@ -0,0 +1,121 @@
...
+
+    if read:
+        # marking as read
+        user_map.read_at = datetime.now(pytz.UTC)
+    else:
+        # marking as unread
+        user_map.read_at = None
+
+    store.save_notification_user_map(user_map)
```
cpennington: Can you get a `NotificationUserMap` from anything but an existing store? Does it make sense to save a map from one store to another? If not, maybe it would be better for the `NotificationUserMap` objects provided by the store to have the store that they will save to baked into them, so that you would just call `user_map.save()`.

On the plus side, this might also mean that you could just return the django-backed user_map object directly, since it already implements that interface.

cpennington:
Broadly, what are the constraints on the store? This repo is going to a lot of effort to allow multiple implementations of the backing store, but I wonder whether it might make more sense to simply constrain the backing store to something we know will work.

Have you done a back-of-the-envelope calculation around cardinality of the various tables, based on how many users and events we could expect to see on edx.org (and possibly for comparison, on McKinsey academy)?

edx_notifications/data.py
```diff
@@ -0,0 +1,172 @@
...
+
+    # the human readible string of the name of the notification
+    display_name = StringField()
+
+    # the human readible string that describes the notification
+    display_description = StringField()
+
+    # default delivery channel for this type
+    # None = no default
+    default_channel = RelatedObjectField(NotificationChannel)
```
cpennington: Does this data need to be stored in the database? Or could the notification types simply be objects that are registered in-process during system startup (with no data storage needed, akin to how XBlocks are loaded)?

edx_notifications/data.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+
+    # the internal name of the channel
+    name = StringField()
+
+    # the human readible name of the channel
+    display_name = StringField()
+
+    # the human readible description of the channel
+    description_name = StringField()
```
cpennington: Likewise, do these need to be stored in a database? Or are they just configured and resident in memory?

edx_notifications/data.py
```diff
@@ -0,0 +1,172 @@
...
+    user_context = DictField()
+
+
+class NotificationUserGroup(BaseDataObject):
+    """
+    A grouping of users into a group, e.g. "cohort", "course enrollments", etc.
+    Note that in order to decouple the notification subsystem
+    from any existing edx-platform dependencies, we need to
+    have generic, abstract grouping mechanisms
+    """
```
cpennington: This isn't used anywhere. Is this just prep for a later feature?

edx_notifications/base_data.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """
+        Initializer which takes in the type this field
+        should be set it is set
+        """
+
+        self._expected_type = args[0]
+        self._default = kwargs.get('default', None)
+        self._data = WeakKeyDictionary()
```
cpennington: Why store the data for each instance on the field object, rather than storing it on the instance objects?

edx_notifications/base_data.py
```diff
@@ -0,0 +1,286 @@
...
+import copy
+import inspect
+import dateutil.parser
+
+from datetime import datetime
+
+from weakref import WeakKeyDictionary
+
+
+class TypedField(object):
```
cpennington: Approach-wise, if there's only ever a single store per-process, would it be better to force each store to provide the relevant fields, and then just use the store's built in field types? That way you wouldn't need to convert between the store and the base data object. Instead, the store would provide the base data objects themselves.

chrisndodge:
Thanks @cpennington. I won't have a chance to answer questions/address issues until Monday.

cpennington:
@chrisndodge no problem. I'm don't feel like I'm done yet, so lets just make it an ongoing conversation.

cpennington:
I've copied all of my comments from this thread over to #12.

edx_notifications/stores/store.py
chrisndodge: done

edx_notifications/stores/store.py
chrisndodge: done

PR #4, opened 2015-01-23
Make long strings easier to read

singingwolfboy:
@chrisndodge, this seems more readable to me. Do you agree?


chrisndodge:
Thanks @singingwolfboy , I'm ripping up some of that code now, but I'll adopt your suggestions once the dust settles.

chrisndodge:
he he, we're stepping on the same areas of code. Need to rebase. Can you hold off for a bit?

PR #5, opened 2015-01-23
Updated pylint

singingwolfboy:
This PR updates pylint to 1.4.1. I also replaced the existing `.pylintrc` file with the output from running `pylint --generate-rcfile`, lightly customized.


chrisndodge:
build is failing because DjangoORM generates that .objects attribute. I believe I had listed that as a 'generated-name' or something like that in the pylintrc file already.

nedbat:
You might like to use edx_lint, the way edx-platform now does, to avoid every repo having their own slightly different pylintrc file.  It also adds a few custom checkers for things which have bitten us in the past.

PR #7, opened 2015-01-29
Cdodge/rename

chrisndodge:


chrisndodge:
@muhhshoaib @afzaledx based on your feedback, I've renamed the django app from 'notifications' to 'edx_notification'. It's a shame because that 'notifications' app in the Django Wiki isn't used at all, but - since it's already done south migrations - I think we cannot deal with the name collisions.

Do you want to try to pull this hash into the LMS and see if it resolves your work?

muhhshoaib:
@chrisndodge I have tried this hash and it is working fine but there is a slight problem when i run the migrations for edx_notifications, i got an error saying "_mysql_exceptions.Warning: Data truncated for column 'name' at row 14". I will send you the full traceback at hipchat. The tables were created just fine and there is a new entry in the southmigrationhistory table as well but i can't figure out why this error displays when i run the migration.

chrisndodge:
Thanks for trying. Not sure what to say about that warning. Seems like the table already had some data in there if it is complaining about row 14?!?

BTW, I just pushed another modification to keep the table name prefixes more in line with Django apps. Before the tables were prefixed "notifications_xxxx", now it'll be "edx_notifications_xxxx". I'm not happy with all that extra name noise, but it's probably best to use standard Django practices.

So, that said, if you pick up these changes you'll have to delete the migration history row and DROP TABLE all the existing notification_xxxx table names in the LMS database.

PR #9, opened 2015-01-29
renamed edx_notifications.store to edx_notifications.stores

chrisndodge:


chrisndodge:
@muhhshoaib note, I've also changed the path edx_notifications.store -> edx_notifications.stores, so you'll have to update your settings. I've updated the README.md.

PR #12, opened 2015-02-02
For ongoing reviews/comments

chrisndodge:


chrisndodge:
@cpennington @ormsbee opened new PR from master to 'empty', so that it keeps up to date with changes. I'll keep other PR open since you have comments on that.

chrisndodge:
@martynjames @mattdrayer FYI, in case you are interested in tagging along.

edx_notifications/stores/store.py
```diff
@@ -0,0 +1,152 @@
...
+
+        ARGS:
+            - msg: an instance of NotificationMessage. If the 'id' field is
+                   set by the caller, then it is assumed to be an update
+                   operation
+
+        RETURNS: type NotificationMessage
+        """
+
+    def save_notification_user_map(self, user_map):
```
cpennington: From #3: This should probably be abstract.

edx_notifications/stores/store.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        Create or Update the mapping of a user to a notification.
+        """
+
+    @abc.abstractmethod
+    def get_notification_type(self, name):
+        """
+        This returns a NotificationType object.
+        """
```
cpennington: From #3: It would be good to raise NotImplementError in these methods, so that if someone super chains up to them, they'll actually fail, rather than just returning None.

I tend to follow the guidelines that abstract methods that change state should have empty base methods (no state change), and that abstract methods that return things should raise errors (no way to return an unknown empty thing from the base). And I try not to have methods that do both. =)

edx_notifications/api/consumer.py
```diff
@@ -0,0 +1,121 @@
...
+        user_id: The id of the user to query
+        filters: A dict containing the following optional keys
+            - 'namespace': what namespace to search (defuault None)
+            - 'read': Whether to return read notifications (default True)
+            - 'unread': Whether to return unread notifications (default True)
+            - 'type_name': what msg_types to count
+
+    NOTE: If the 'msg_id' filter is passed in, a ValueError will be raised as
+    this is not a valid combination
+    """
```
cpennington: From #3: Would it be better to just have these as optional arguments to this function? Or to have a Filter class that defines the set of valid filters?

edx_notifications/api/consumer.py
```diff
@@ -0,0 +1,121 @@
...
+    Returns a list of NotificationUserMap for the passed in user. The list will be
+    ordered by most recent first
+
+    ARGS:
+        - user_id: The id of the user
+        - filters: a dict containing
+            - msg_id: a particular notification id to retrieve
+                ( if this is used in conjuction with other filters, it may
+                  result in an empty set being returned )
```
cpennington: From #3: `msg_id` seems to always be incompatible with other filters, right? If you're querying for a specific message, it doesn't make sense to query anything else at the same time.

edx_notifications/api/consumer.py
```diff
@@ -0,0 +1,121 @@
...
+        # at least with SQL backends. However, for future no-SQL backends
+        # this is conceptually a possibility, so good to double check
+        err_msg = (
+            'There should be at most 1 NotificationUserMap items found for '
+            'msg_id {msg_id} and user_id {user_id}. {cnt} were found!'
+        ).format(msg_id=msg_id, user_id=user_id, cnt=len(notifications))
+
+        raise ItemIntegrityError(err_msg)
+
+    user_map = notifications[0]
```
cpennington: From #3: `user_map` is a confusing name for the results of `get_notifications_for_user`. Is there a better name for that concept? Perhaps `UserNotificationMessage` (in contrast to `NotificationMessage` for the raw message)?

edx_notifications/api/consumer.py
```diff
@@ -0,0 +1,121 @@
...
+
+    if read:
+        # marking as read
+        user_map.read_at = datetime.now(pytz.UTC)
+    else:
+        # marking as unread
+        user_map.read_at = None
+
+    store.save_notification_user_map(user_map)
```
cpennington: From #3: Can you get a `NotificationUserMap` from anything but an existing store? Does it make sense to save a map from one store to another? If not, maybe it would be better for the `NotificationUserMap` objects provided by the store to have the store that they will save to baked into them, so that you would just call `user_map.save()`.

On the plus side, this might also mean that you could just return the django-backed user_map object directly, since it already implements that interface.

cpennington:
Broadly, what are the constraints on the store? This repo is going to a lot of effort to allow multiple implementations of the backing store, but I wonder whether it might make more sense to simply constrain the backing store to something we know will work.

Have you done a back-of-the-envelope calculation around cardinality of the various tables, based on how many users and events we could expect to see on edx.org (and possibly for comparison, on McKinsey academy)?

edx_notifications/data.py
```diff
@@ -0,0 +1,172 @@
...
+
+    # the human readible string of the name of the notification
+    display_name = StringField()
+
+    # the human readible string that describes the notification
+    display_description = StringField()
+
+    # default delivery channel for this type
+    # None = no default
+    default_channel = RelatedObjectField(NotificationChannel)
```
cpennington: Does this data need to be stored in the database? Or could the notification types simply be objects that are registered in-process during system startup (with no data storage needed, akin to how XBlocks are loaded)?

edx_notifications/data.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+
+    # the internal name of the channel
+    name = StringField()
+
+    # the human readible name of the channel
+    display_name = StringField()
+
+    # the human readible description of the channel
+    description_name = StringField()
```
cpennington: From #3: Likewise, do these need to be stored in a database? Or are they just configured and resident in memory?

edx_notifications/data.py
```diff
@@ -0,0 +1,172 @@
...
+    user_context = DictField()
+
+
+class NotificationUserGroup(BaseDataObject):
+    """
+    A grouping of users into a group, e.g. "cohort", "course enrollments", etc.
+    Note that in order to decouple the notification subsystem
+    from any existing edx-platform dependencies, we need to
+    have generic, abstract grouping mechanisms
+    """
```
cpennington: From #3: This isn't used anywhere. Is this just prep for a later feature?

edx_notifications/base_data.py
```diff
@@ -0,0 +1,287 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """
+        Initializer which takes in the type this field
+        should be set it is set
+        """
+
+        self._expected_type = [args[0]] if not isinstance(args[0], list) else args[0]
+        self._default = kwargs.get('default', None)
+        self._data = WeakKeyDictionary()
```
cpennington: From #3: Why store the data for each instance on the field object, rather than storing it on the instance objects?

edx_notifications/base_data.py
```diff
@@ -0,0 +1,287 @@
...
+import copy
+import inspect
+import dateutil.parser
+
+from datetime import datetime
+
+from weakref import WeakKeyDictionary
+
+
+class TypedField(object):
```
cpennington: Approach-wise, if there's only ever a single store per-process, would it be better to force each store to provide the relevant fields, and then just use the store's built in field types? That way you wouldn't need to convert between the store and the base data object. Instead, the store would provide the base data objects themselves.

edx_notifications/lib/consumer.py
```diff
@@ -0,0 +1,121 @@
...
+        user_id: The id of the user to query
+        filters: A dict containing the following optional keys
+            - 'namespace': what namespace to search (defuault None)
+            - 'read': Whether to return read notifications (default True)
+            - 'unread': Whether to return unread notifications (default True)
+            - 'type_name': what msg_types to count
+
+    NOTE: If the 'msg_id' filter is passed in, a ValueError will be raised as
+    this is not a valid combination
+    """
```
cpennington: This got lost in the rename: Would it be better to just have these as optional arguments to this function? Or to have a `Filter` class that defines the set of valid filters?

edx_notifications/lib/consumer.py
```diff
@@ -0,0 +1,121 @@
...
+    Returns a list of NotificationUserMap for the passed in user. The list will be
+    ordered by most recent first
+
+    ARGS:
+        - user_id: The id of the user
+        - filters: a dict containing
+            - msg_id: a particular notification id to retrieve
+                ( if this is used in conjuction with other filters, it may
+                  result in an empty set being returned )
```
cpennington: This got lost in the rename: `msg_id` seems to always be incompatible with other filters, right? If you're querying for a specific message, it doesn't make sense to query anything else at the same time.

edx_notifications/lib/consumer.py
```diff
@@ -0,0 +1,121 @@
...
+        # at least with SQL backends. However, for future no-SQL backends
+        # this is conceptually a possibility, so good to double check
+        err_msg = (
+            'There should be at most 1 NotificationUserMap items found for '
+            'msg_id {msg_id} and user_id {user_id}. {cnt} were found!'
+        ).format(msg_id=msg_id, user_id=user_id, cnt=len(notifications))
+
+        raise ItemIntegrityError(err_msg)
+
+    user_map = notifications[0]
```
cpennington: This got lost in the rename: `user_map` is a confusing name for the results of `get_notifications_for_user`. Is there a better name for that concept? Perhaps `UserNotificationMessage` (in contrast to `NotificationMessage` for the raw message)?

edx_notifications/lib/consumer.py
```diff
@@ -0,0 +1,121 @@
...
+
+    if read:
+        # marking as read
+        user_map.read_at = datetime.now(pytz.UTC)
+    else:
+        # marking as unread
+        user_map.read_at = None
+
+    store.save_notification_user_map(user_map)
```
cpennington: This got lost in the rename: Can you get a `NotificationUserMap` from anything but an existing store? Does it make sense to save a map from one store to another? If not, maybe it would be better for the `NotificationUserMap` objects provided by the store to have the store that they will save to baked into them, so that you would just call `user_map.save()`.

On the plus side, this might also mean that you could just return the django-backed user_map object directly, since it already implements that interface.

edx_notifications/base_data.py
```diff
@@ -0,0 +1,287 @@
...
+        super(StringField, self).__init__(unicode, kwargs)
+
+    def __set__(self, instance, value):
+        """
+        Check to see if we were passed a str type, and if so
+        coerce it into a unicode
+        """
+
+        if isinstance(value, str):
+            super(StringField, self).__set__(instance, unicode(value))
```
cpennington: This doesn't make much sense if you don't know the encoding of the string, does it?

edx_notifications/base_data.py
```diff
@@ -0,0 +1,287 @@
...
+        else:
+            super(StringField, self).__set__(instance, value)
```
cpennington: Nit: you could reduce some of the duplication by updating value in an if-statement, and then always calling `super` the same way.

edx_notifications/base_data.py
```diff
@@ -0,0 +1,287 @@
...
+
+    _expected_type = None
+
+    def __init__(self, *args, **kwargs):
+        """
+        Initializer which takes in the type this field
+        should be set it is set
+        """
+
+        self._expected_type = [args[0]] if not isinstance(args[0], list) else args[0]
```
cpennington: If you expect `expected_type` to be the first argument, why not just include it in the method signature?

edx_notifications/base_data.py
```diff
@@ -0,0 +1,287 @@
...
+            super(StringField, self).__set__(instance, value)
+
+
+class IntegerField(TypedField):
+    """
+    Specialized subclass of TypedField(int) as a convienence
+    """
+
+    def __init__(self, **kwargs):
+        super(IntegerField, self).__init__([int, long], kwargs)
```
cpennington: Seeing all of these one-line init functions makes me feel like the expected type should be a class attribute, rather than something passed in the constructor. Is it important to be able to construct new `TypedField`s on the fly, without a new subclass?

edx_notifications/base_data.py
```diff
@@ -0,0 +1,287 @@
...
+class DictField(TypedField):
+    """
+    Specialized subclass of TypedField(dict) as a convienence
+    """
+
+    def __init__(self, **kwargs):
+        super(DictField, self).__init__(dict, kwargs)
+
+    @classmethod
+    def to_json(cls, data):
```
cpennington: This is the only field class with `to_json` and `from_json` defined. Are they part of the `TypedField` interface?

edx_notifications/base_data.py
```diff
@@ -0,0 +1,287 @@
...
+        if not data:
+            return None
+
+        _dict = copy.deepcopy(data)
+
+        for key, value in _dict.iteritems():
+            if isinstance(value, datetime):
+                _dict[key] = value.isoformat()
+
+        return json.dumps(_dict)
```
cpennington: You're json-dumping the dict, which means that you're immediately throwing away the data you just deepcopied. This seems like a potential nasty perf hit. It would be better to specify a custom JSON encoder that knows how to convert `datetime` objects to isoformat.

edx_notifications/base_data.py
```diff
@@ -0,0 +1,287 @@
...
+            return None
+
+        _dict = json.loads(value)
+
+        for key, value in _dict.iteritems():
+            if isinstance(value, basestring):
+                # This could be a datetime posing as a ISO8601 formatted string
+                # we so have to apply some heuristics here
+                # to see if we want to even attempt
+                if value.count('-') == 2 and value.count(':') == 2 and value.count('T') == 1:
```
cpennington: Rather than relying on string introspection, maybe dates should be serialized as a dictionary with a `type` key, or something like that?

edx_notifications/base_data.py
```diff
@@ -0,0 +1,287 @@
...
+
+class BaseDataObject(object):
+    """
+    A base class for all Notification Data Objects
+    """
+
+    id = IntegerField(default=None)  # pylint: disable=invalid-name
+
+    _is_loaded = False
+    _is_dirty = False  # Returns if this object has been modified after initialization
```
cpennington: `_is_dirty` is never used. Also, from experience, detecting `dirty` when you have mutable fields is a pain.

edx_notifications/base_data.py
```diff
@@ -0,0 +1,287 @@
...
+        is a problem.
+        """
+        pass  # this intentionally does nothing
+
+
+class RelatedObjectField(TypedField):
+    """
+    This field is a related object that is joined to the containing object,
+    for example a foreign key. The related object must be of type BaseDataObject
+    """
```
cpennington: Is this a one-to-one relation? One-to-many? Many-to-many?

edx_notifications/base_renderer.py
```diff
@@ -0,0 +1,70 @@
...
+        if the NotificationChannel *must* have a subject line, for example
+        email-based delivery channels.
+
+        If the requested language is not supported then subclasses should
+        throw a NotificationLanguageNotSupported exception. The calling code
+        should trap that and try with a different language
+        """
+
+        # Base implementation will say it cannot render a Subject
+        return None
```
cpennington: `return None` is the same as having an empty body for the method. Perhaps you meant `raise NotImplementError`?

cpennington: Oh, nevermind. I hadn't read the full docstring. This is fine as-is.

edx_notifications/base_renderer.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @abc.abstractmethod
+    def render_body(self, msg, render_format, lang):
+        """ for
```
cpennington: Typo?

edx_notifications/channels/channel.py
```diff
@@ -0,0 +1,206 @@
...
+    be called in normal runtimes
+
+    NOTE: When we switch over to gevent support, we should use some
+    locking techniques in this area to prevent concurrent execution
+    """
+
+    global _CHANNEL_PROVIDERS  # pylint: disable=global-statement
+    global _CHANNEL_PROVIDERS_TYPE_MAPS  # pylint: disable=global-statement
+
+    _CHANNEL_PROVIDERS = {}
```
cpennington: You could use `_CHANNEL_PROVIDERS.clear()`

edx_notifications/channels/channel.py
```diff
@@ -0,0 +1,206 @@
...
+    _CHANNEL_PROVIDERS_TYPE_MAPS = {}
+
+
+class BaseNotificationChannelProvider(object):
```
cpennington: This could perhaps be written shorter by just using a `namedtuple` as a baseclass.

cpennington: Although I guess you'd still have to do some overrides to allow the arguments to be optional.

edx_notifications/channels/tests/test_channel.py
```diff
@@ -0,0 +1,196 @@
...
+class ChannelTests(TestCase):
+    """
+    Tests for channel.py
+    """
+
+    def setUp(self):
+        """
+        Harnessing
+        """
+        reset_notification_channels()
```
cpennington: Would it make sense to call this as a `cleanUp` action as well?

edx_notifications/data.py
cpennington: Aren't these the same channels that are configured in the settings file? Why are they both configured in settings and also stored in the database?

edx_notifications/data.py
```diff
@@ -0,0 +1,172 @@
...
+    default_channel = RelatedObjectField(NotificationChannel)
+
+
+class NotificationMessage(BaseDataObject):
+    """
+    The basic Notification Message
+    """
+
+    # instance of NotificationMessageType, None = unloaded
+    msg_type = RelatedObjectField(NotificationType)
```
cpennington: What does `unloaded` mean in this context?

edx_notifications/base_data.py
```diff
@@ -0,0 +1,287 @@
...
+        return self._data.get(instance, self._default)
+
+    def __set__(self, instance, value):
+        """
+        Descriptor setter. Be sure to enforce type on setting. But None is allowed.
+        """
+
+        value_type = type(value)
+
+        if value and type(value) not in self._expected_type:
```
cpennington: Use `isinstance`, rather than `type(value) in self._expected_type`, otherwise you won't handle type subclasses.

edx_notifications/lib/consumer.py
```diff
@@ -0,0 +1,121 @@
...
+            - unread: Whether to return unread notifications (default True)
+            - type_name: which type to return
+        - options: a dict containing some optional parameters
+            - limit: max number to return (up to some system defined max)
+            - offset: offset into the list, to implement paging
+    """
+
+    # make sure user_id is an integer
+    if not isinstance(user_id, int):
+        raise TypeError("user_id must be an integer")
```
cpennington: Optional: we've been using pycontracts (https://andreacensi.github.io/contracts/) for this sort of thing in edx-platform. Is that something you'd want to adopt here?

edx_notifications/lib/tests/test_lib.py
```diff
@@ -0,0 +1,191 @@
...
+        Go through and set up a notification and publish it
+        """
+
+        msg = NotificationMessage(
+            namespace='test-runner',
+            msg_type=self.msg_type,
+            payload={
+                'foo': 'bar'
+            }
+        )
```
cpennington: If message creation were a method on the `NotificationType`, then it could enforce that the required and optional arguments in the payload were passed correctly.

cpennington: So the call would look like:

    self.msg_type.create(namespace='test-runner', foo='bar')

edx_notifications/lib/tests/test_lib.py
```diff
@@ -0,0 +1,191 @@
...
+        self.assertTrue(isinstance(notifications, list))
+        self.assertEqual(len(notifications), 1)
+        self.assertTrue(isinstance(notifications[0], NotificationUserMap))
+
+        read_user_map = notifications[0]
+        self.assertEqual(read_user_map.user_id, self.test_user_id)
+        self.assertIsNone(read_user_map.read_at)  # should be unread
+
+        self.assertEqual(read_user_map, sent_user_map)
+        self.assertEqual(read_user_map.msg, sent_user_map.msg)
```
cpennington: Generally, I tend to prefer test methods that are more self-contained. Couldn't you move the various unhappy-path branches into their own specific test methods?

cpennington:
Is the expectation that this will be running as an independent process? (I note the `wsgi.py` file). If so, what version of django were you planning on running under? If possible, can we run it under Django 1.7? (I noticed `user_passes_test` defined locally, which seems like you were planning on Django 1.4).

edx_notifications/stores/sql/models.py
```diff
@@ -0,0 +1,256 @@
...
+        obj = SQLNotificationType()
+        obj.load_from_data_object(msg_type)
+        return obj
+
+    def load_from_data_object(self, msg_type):
+        """
+        Hydrate ourselves from a passed in user_map
+        """
+
+        self.name = msg_type.name  # pylint: disable=attribute-defined-outside-init
```
cpennington: Why two methods for this? Why not just instantiate the `SQLNotificationType` with a name directly?

edx_notifications/stores/sql/models.py
```diff
@@ -0,0 +1,256 @@
...
+    string_value = models.TextField()
+
+    class Meta(object):
+
+        """
+        ORM metadata about this class
+        """
+        app_label = 'edx_notifications'  # since we have this models.py file not in the root app directory
+        db_table = 'edx_notifications_displaystring'
+        unique_together = (('string_name', 'lang'),)
```
cpennington: What's this model for?

edx_notifications/stores/sql/store_provider.py
```diff
@@ -0,0 +1,283 @@
...
+
+class SQLNotificationStoreProvider(BaseNotificationStoreProvider):
+    """
+    Concrete MySQL implementation of the abstract base class (interface)
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Initializer
+        """
```
cpennington: This could use some documentation about what `kwargs` are expected/allowed.

edx_notifications/stores/sql/store_provider.py
```diff
@@ -0,0 +1,283 @@
...
+
+        # pylint/pep8 seem to complain if defaults are set to empty dicts
+        _options = options if options else {}
+        select_related = _options.get('select_related', True)
+
+        try:
+            if select_related:
+                obj = SQLNotificationMessage.objects.select_related().get(id=msg_id)
+            else:
+                obj = SQLNotificationMessage.objects.get(id=msg_id)
```
cpennington: I think you could write this as:

    query = SQLNotificationMessage.objects.get(id=msg_id)
    if select_related:
        query = query.select_related()

cpennington: Actually, I guess it would need to be:


    query = SQLNotificationMessage.objects
    if select_related:
        query = query.select_related()
    obj = query.get(id=msg_id)

edx_notifications/stores/sql/store_provider.py
```diff
@@ -0,0 +1,283 @@
...
+        it will try to update the object. If it does not exist it will throw an
+        exception.
+
+        If it is created, then the id property will be set on the NotificationMsg and returned
+        """
+
+        if msg.id:
+            try:
+                obj = SQLNotificationMessage.objects.get(id=msg.id)
+                obj.load_from_data_object(msg)
```
cpennington: Ah, I see now why there were two different methods for loading a data object.

edx_notifications/stores/sql/store_provider.py
```diff
@@ -0,0 +1,283 @@
...
+        if select_related:
+            query = query.select_related()
+
+        if msg_id:
+            query = query.filter(msg__id=msg_id)
+
+        if namespace:
+            query = query.filter(msg__namespace=namespace)
+
+        if not (read and unread):
```
cpennington: Should specifying `read` and `unread` be an error?

edx_notifications/stores/sql/store_provider.py
```diff
@@ -0,0 +1,283 @@
...
+        if filters and 'msg_id' in filters:
+            raise ValueError(
+                'It is non-sensical to pass in a msg_id '
+                'filter parameter into an aggregate counting method!'
+            )
+
+        return self._get_notifications_for_user(
+            user_id,
+            filters=filters,
+        ).count()
```
cpennington: Is this `count()` going to play well with the `limit` part of `_get_notifications_for_user`?

cpennington: Seems like it would be worth a test for things larger than the max limit.

edx_notifications/stores/sql/store_provider.py
```diff
@@ -0,0 +1,283 @@
...
+                - type_name: which type to return
+            - options: a dict containing some optional parameters
+                - limit: max number to return (up to some system defined max)
+                - offset: offset into the list, to implement paging
+
+        RETURNS: list   i.e. []
+        """
+
+        _options = copy.deepcopy(options) if options else {}
+        _options['select_related'] = True
```
cpennington: You don't need a `deepcopy` here. You're only modifying the top of the dict, so a regular `.copy()` would be enough.

edx_notifications/stores/sql/store_provider.py
```diff
@@ -0,0 +1,283 @@
...
+
+        query = self._get_notifications_for_user(
+            user_id,
+            filters=filters,
+            options=_options
+        )
+
+        result_set = []
+        for item in query:
+            result_set.append(item.to_data_object())
```
cpennington: `result_set = [item.to_data_object() for item in query]`?

chrisndodge:
@cpennington you asked about the runtime process. Our MVP will be running in-proc. The wsgi.py is here to support test harnessing for UI acceptance tests. Later versions will have this running out-of-proc and using the HTTP REST endpoints which are being written now.

chrisndodge:
@cpennington one general comment. As you observed there are a few models which are not used. This is just sketch work, but perhaps I should remove all known 'scratchpad' stuff for now.

edx_notifications/stores/sql/store_provider.py
```diff
@@ -0,0 +1,283 @@
...
+            try:
+                obj = SQLNotificationUserMap.objects.get(id=user_map.id)
+                obj.load_from_data_object(user_map)
+            except ObjectDoesNotExist:
+                msg = "Could not find SQLNotificationUserMap with ID {_id}".format(_id=user_map.id)
+                raise ItemNotFoundError(msg)
+        else:
+            obj = SQLNotificationUserMap.from_data_object(user_map)
+
+        obj.save()
```
cpennington: Could you use `get_or_create` here, to simplify this code?

edx_notifications/stores/sql/store_provider.py
chrisndodge: read + unread = all

edx_notifications/base_data.py
chrisndodge: how would I store it on the instance object?

edx_notifications/data.py
chrisndodge: Yea, I'm going to remove all known 'scratchpad' stuff.

edx_notifications/data.py
chrisndodge: This was day 1 code, before I wrote up channel.py as it current stands. Like elsewhere noted, I'm going to remove this 'scratchpad' stuff.

chrisndodge:
@cpennington in terms of your question on cardinality, the customer that will be running this first will be in the order of:

- low 100's courses
- 200 to 300 students per course

Notification triggers will primary be sourced from Forum activity, looming due dates, and course announcements. So maybe < 100 Notifications per course per student

edx_notifications/data.py
chrisndodge: I'm debating whether to change this to an in-memory for now since MVP will be running in-proc. I think one of the original appeal to have this go into SQL was to have a ForeignKey constraint on the type_name.

edx_notifications/base_data.py
chrisndodge: Let's talk about this one.

edx_notifications/lib/consumer.py
chrisndodge: I can see your point. I started down this path because of pylint starting to complain about too many optional arguments :-)

That said, I think having a schema'd Filter class make sense.

edx_notifications/lib/consumer.py
chrisndodge: OK, I like UserNotificationMessage. Will do.

edx_notifications/lib/consumer.py
chrisndodge: Let talk about this. As mentioned above, my goal was to make plain-old Data Objects and not expose methods on them. As a result, that makes things more 'verb-y' as you note.

edx_notifications/base_data.py
chrisndodge: Actually I changed the FIelds to now support a list of types. I'm not sure the coercion is necessary any long.

edx_notifications/stores/sql/store_provider.py
cpennington: Ah, I see. Makes sense.

edx_notifications/base_data.py
cpennington: In `__set__`, you have the instance object. You could just do something like: `instance.__data[self] = value`.

edx_notifications/base_data.py
chrisndodge: I think I see what you mean

edx_notifications/base_data.py
chrisndodge: I can add a comment to this affect but this is primarily to support storing dicts as TextFields() in SQL.

cpennington:
so, probably 100 courses * 300 students * 100 notifications ~= 3M NotificationUserMap objects?

edx_notifications/base_data.py
chrisndodge: I'll look into that. Any prior art that you know of?

cpennington:
Is there a plan for message expiry? Given the large number of notification objects for that relatively small instance, it seems like it's going to be important to get rid of messages as soon as possible.

edx_notifications/base_data.py
chrisndodge: do you mean a string "magic" prefix?

edx_notifications/base_data.py
chrisndodge: yea, this is unused scratchpad stuff. Will remove until it is needed.

edx_notifications/base_data.py
chrisndodge: one-to-one (at least so far). We're already pushing the boundaries for this getting very ORM-like

edx_notifications/lib/consumer.py
chrisndodge: Thanks for the tip. Hadn't heard of this before.

edx_notifications/lib/tests/test_lib.py
chrisndodge: I guess this goes back to if it is preferable to have pure data objects, or have these data objects also have methods on them.

edx_notifications/lib/tests/test_lib.py
chrisndodge: OK, I see your point. Often there's some data setup which has to happen before the assertions happen, so it ends up with duplicative code. But I can separate these out.

edx_notifications/stores/sql/models.py
chrisndodge: scratch pad stuff, but - in essence - I was thinking through what it might mean to have display strings in a database which might need to support multiple languages. But I'm not sure this will actually be necessary.

edx_notifications/stores/sql/store_provider.py
chrisndodge: yea, sometimes we already have a Django ORM object

chrisndodge:
@cpennington I was planning to write a Django management command to purge messages and have this run periodically via recurring schedule Jenkins Job.

edx_notifications/tests/test_data.py
```diff
@@ -0,0 +1,27 @@
+"""
+Unit tests to exercise code implemented in data.py
+"""
+
+from django.test import TestCase
+
+from django.core.exceptions import ValidationError
+
+from edx_notifications.data import (  # pylint: disable=unused-import
```
cpennington: Why the pylint disable?

pylintrc
```diff
@@ -0,0 +1,260 @@
...
+# can either give multiple identifiers separated by comma (,) or put this
+# option multiple times (only on the command line, not in the configuration
+# file where it should appear only once).You can also use "--disable=all" to
+# disable everything first and then reenable specific checks. For example, if
+# you want to run only the similarities checker, you can use "--disable=all
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use"--disable=all --enable=classes
+# --disable=W"
+# I0011     locally-disabled   (module-level pylint overrides)
+disable=I0011,W0232,too-few-public-methods,abstract-class-little-used,abstract-class-not-used,too-many-public-methods,no-self-use,too-many-instance-attributes
```
cpennington: It would be good to always use the named errors.

run_tests
```diff
@@ -0,0 +1,15 @@
+ECHO 'Beginning Test Run...'
```
cpennington: We've been using `Makefile`s for this kind of thing recently.

edx_notifications/stores/store.py
chrisndodge: done

edx_notifications/lib/tests/test_lib.py
cpennington: I tend to lean towards a cleaner, more informative API where possible. I think having explicit parameters will help with that.

edx_notifications/base_data.py
```diff
@@ -0,0 +1,287 @@
...
+        return self._data.get(instance, self._default)
+
+    def __set__(self, instance, value):
+        """
+        Descriptor setter. Be sure to enforce type on setting. But None is allowed.
+        """
+
+        value_type = type(value)
+
+        if value and type(value) not in self._expected_types:
```
cpennington: Use `isinstance`, rather than `type(value) in self._expected_type`, otherwise you won't handle type subclasses.

edx_notifications/base_data.py
cpennington: If its particular to the SQL implementation, why does it need to be in the general base class?

edx_notifications/data.py
cpennington: Yeah, I could see how the `ForeignKey` might be nice. It's unfortunate to have to keep saving things to the database.

chrisndodge:
@nasthagiri tagging you as an FYI since your team might be building on top of this next quarter. @cpennington is doing ongoing code/design reviews, but I think this is enough code here for us to cc: you on this so you can choose to follow along as you wish. This PR is basically a "faux-PR" which reflects all ongoing work. It's just helpful to have an area for people to scratch comments on.

In short, I would anticipate the Mobile team writing a subclass of NotificationChannel in order to render and dispatch the notification msg via the various mobile Push Notifications frameworks. Next Sprint we'll be taking on some of the "Render Notifications in a Browser" stories, so some of the planned designed around rendering will be fleshed out in the next two weeks.

When I get a breather, I'll probably start building up some technical documentation in this Repo's wiki.

FYI, @smagoun @karadhya

chrisndodge:
@cpennington re: using Django 1.7 when 'running as a separate service'. Yea that should work, however, does south still work with 1.7? I say this because MVP will be running in-proc, so I'd like to make sure that the existing migrations defined in the repo still work (since that will be running 1.4).

So, what point versions should I look into for 1.7 as well as 1.6 (in case there's an incompatibility between South and 1.7)? I'd at least like to get to 1.6 in order to at least get the SQL connection pooling which would be very helpful for the REST API endpoints since those are very short lived requests.

cpennington:
South does work with 1.7. Eventually, I'd like to just be using the new 1.7 migration framework, but there is a path to switch from South to that, which we can follow once it's actually running out-of-proc.

I wouldn't use 1.6 at all. I'd just go with 1.4 and 1.7. It may not be possible (or feasible) to write it such that both 1.4 and 1.7 work at the same time, though.

chrisndodge:
@andy-armstrong as we discussed on HipChat, I just put together a single initial sketch at a Backbone application/models/views/templates for the Notifications. My goal is to make as much as possible re-useable between McKinsey Academy and Open edX/edx.org. Hopefully - depending on what Product would like to see on Open edX/edx.org - we'll just need to change the Underscore templates but the models and - most of the views - should be the same.

I'd love to get high-level feedback as this will be my first Backbone application. I will be passing on the real development to a remote team next week, and I'd like to make sure we're heading in the right direction and using good patterns.

Basically, can you take a look at /edx_notifications/server/web (high contain the Backbone/Underscore stuff) as well as /edx_notifications/testserver, which is a development & test website which hosts the Backbone UI widgets.

Thank you

edx_notifications/lib/consumer.py
chrisndodge: done

edx_notifications/server/web/static/edx_notifications/js/models/counter_icon_model.js
```diff
@@ -0,0 +1,11 @@
+var CounterIconModel = Backbone.Model.extend({
```
andy-armstrong: It would be best if your JS code could use RequireJS. Look at how content search used it, for example:

https://github.com/edx/edx-platform/blob/master/lms/static/js/search/models/search_result.js

If not, then our best practice is to put everything into the edX namespace. Here's a simple example:

https://github.com/edx/edx-platform/blob/master/lms/static/js/student_account/models/LoginModel.js

edx_notifications/server/web/static/edx_notifications/js/models/counter_icon_model.js
```diff
@@ -0,0 +1,11 @@
...
+    /* model for the Notification Icon which will display a counter of unread messages */
+
+    defaults: {
+        /* start with unknown number of unread notifications */
+        'count': null
+    },
+
+    /* REST API endpoint to use to get authenticated user's notification count */
+    url: "/api/edx_notifications/v1/consumer/notifications/count?unread=True&read=False"
```
andy-armstrong: Our best practice is to pass the URL down from the server, rather than having it hard-coded on the client. Maybe that is less important going forward as we have very cleanly namespaced URLs like this one.

edx_notifications/server/web/static/edx_notifications/js/vendor/dev/backbone.js
```diff
@@ -0,0 +1,1608 @@
+//     Backbone.js 1.1.2
```
andy-armstrong: It seems a shame to have another copy of Backbone (and other third-party libraries) here. Could you use RequireJS to declare it as a dependency and just assume that it will always be deployed somewhere that already has Backbone installed?

edx_notifications/server/web/static/edx_notifications/js/models/counter_icon_model.js
chrisndodge: Yea, I was thinking about this. Is it typically rendered onto to some 'data-' attribute?

edx_notifications/server/web/static/edx_notifications/js/views/counter_icon_view.js
```diff
@@ -0,0 +1,66 @@
...
+    modelChanged: function() {
+        this.render();
+    },
+
+    render: function () {
+        if (!this.fetched_template) {
+            var self = this;
+
+            /* load Underscore template asynchronously on first load */
+            $.get("/static/edx_notifications/templates/underscore/notification_icon.html")
```
andy-armstrong: Our best practice (that we haven't managed to start using yet!) is to use RequireJS.text to load Underscore templates:

https://github.com/requirejs/text

The beauty of this is that it works with RequireJS Optimizer so that the templates themselves get included in the single JS file that is minified for the entire page.

edx_notifications/server/web/static/edx_notifications/js/models/counter_icon_model.js
chrisndodge: Thanks, looking at search_result.js, I'm not sure about the syntax. I'll talk to @martynjames or you tomorrow when we are back in the office.

edx_notifications/server/web/static/edx_notifications/js/views/notification_pane_view.js
```diff
@@ -0,0 +1,54 @@
...
+
+            /* load Underscore template asynchronously on first load */
+            $.get("/static/edx_notifications/templates/underscore/notification_pane.html")
+                .done(function(template_data) {
+
+                    /* convert to Underscore template object */
+                    self.fetched_template = _.template(template_data);
+
+                    /* now render template with our model */
+                    self.$el.html(
```
andy-armstrong: Shouldn't this render be outside of the caching logic? It looks like it will only render the first time...

If you do decide to roll your own template loading code, I recommend moving it into a helper method that can be shared across your various files. As it stands, the rendering logic is hard to understand as it is obscured by all the template handling.

edx_notifications/server/web/static/edx_notifications/js/views/counter_icon_view.js
```diff
@@ -0,0 +1,66 @@
+var CounterIconView = Backbone.View.extend({
```
andy-armstrong: Please add Jasmine tests for all the JS code.

chrisndodge: Yes, we have a story in this sprint to add the Jasmine (and BokChoy) test harness in edx-notifications. Right now we are interested in generic high-level feedback on the Backbone patterns we are setting up.

edx_notifications/server/web/static/edx_notifications/js/views/counter_icon_view.js
chrisndodge: thx. I might need to sit down with you tomorrow about this recommendation.

edx_notifications/server/web/templates/django/notifications_widget.html
```diff
@@ -0,0 +1,31 @@
...
+<script src="{{ STATIC_URL }}edx_notifications/js/views/counter_icon_view.js"></script>
+<script src="{{ STATIC_URL }}edx_notifications/js/views/notification_pane_view.js"></script>
+
+
+<script type="text/javascript">
+    /* start up the root Notification view, which will bootstrap the whole front-end */
+    var counter_icon = new CounterIconView({
+        el: $("#edx_notification_container"),
+        model: new CounterIconModel(),
+        pane_el: $("#edx_notification_pane")
```
andy-armstrong: Is there a reason why the notification view has a separation between the container and the pane? I think it would be simpler if the view had a single element that contained both. This would then allow the view to be created without an ``el`` and then appended to a parent view wherever it was needed. I prefer using this approach for views in general because you can safely remove the view and it removes its own element, rather than an element from the DOM that was passed in. It should also make it a little easier to unit test as it doesn't require the two divs to be added as fixtures.

edx_notifications/server/web/static/edx_notifications/js/views/counter_icon_view.js
andy-armstrong: That sounds good. In the meantime, there's a little information about it here:

https://openedx.atlassian.net/wiki/display/TNL/How+to+add+a+new+page+to+Studio

edx_notifications/server/web/static/edx_notifications/js/views/notification_pane_view.js
```diff
@@ -0,0 +1,54 @@
...
+    },
+
+    hydrate: function() {
+        /* This function will load the bound collection */
+
+        /* add and remove a class when we do the initial loading */
+        /* we might - at some point - add a visual element to the */
+        /* loading, like a spinner */
+        var self = this;
+        self.$el.addClass('loading');
```
andy-armstrong: In Studio we use the class 'ui-loading' so we might want to standardize on that:

https://github.com/edx/edx-platform/blob/master/cms/static/js/views/utils/view_utils.js#L31

edx_notifications/server/web/templates/django/notifications_widget.html
chrisndodge: yea, let me sit down with you with the mockups. The notification icon will be a permanent element in a header region and then the pane is an overlay which will appear over the rest of the page when the notification icon is clicked. I wasn't sure whether it'd be more easy to get the overlay behavior by having it as a separate element or not.

edx_notifications/server/web/static/edx_notifications/js/models/counter_icon_model.js
andy-armstrong: Yes, that's one way that we do it. The other way is to pass it as a parameter to the root view as long as the view is constructed by JS output from a Mako template.

edx_notifications/server/web/templates/django/notifications_widget.html
```diff
@@ -0,0 +1,31 @@
...
+<script src="{{ STATIC_URL }}edx_notifications/js/models/user_notification_model.js"></script>
+<script src="{{ STATIC_URL }}edx_notifications/js/models/notification_collection_model.js"></script>
+
+<script src="{{ STATIC_URL }}edx_notifications/js/views/counter_icon_view.js"></script>
+<script src="{{ STATIC_URL }}edx_notifications/js/views/notification_pane_view.js"></script>
+
+
+<script type="text/javascript">
+    /* start up the root Notification view, which will bootstrap the whole front-end */
+    var counter_icon = new CounterIconView({
```
andy-armstrong: In order to use RequireJS Optimizer, we are using a pattern now where Mako templates include a single JS file (which we typically call a factory) which knows how to construct the models, views etc needed. Even if you don't use RequireJS yet, it would probably be good to follow this pattern so that we can adopt it in the future.

See this wiki page for more details:

https://openedx.atlassian.net/wiki/display/TNL/How+to+add+a+new+page+to+Studio

edx_notifications/stores/sql/store_provider.py
chrisndodge: ah yea, good catch. I took me awhile to figure out what you meant, but I see the enforced limit even if no OPTIONS are passed in.

edx_notifications/base_data.py
chrisndodge: done

edx_notifications/channels/tests/test_channel.py
chrisndodge: done

edx_notifications/lib/consumer.py
chrisndodge: done, thanks for the pointer. I like the contract pattern

chrisndodge:
@cpennington I believe I got through all of your feedback, minus the 'maybe we should put methods on the DataObjects'. @smagoun would like to get ongoing checkpoint +1's so we can move our JIRA stories over to done. Obviously this is not a final review, but we'd like to strike a balance between keeping the JIRA book-keeping moving along while trying to avoid "big surprises" when the projects moves into final review in a few weeks' time.

edx_notifications/data.py
```diff
@@ -0,0 +1,169 @@
...
+    # the human readible string that describes the notification
+    display_description = StringField()
+
+    # default delivery channel for this type
+    # None = no default
+    default_channel = RelatedObjectField(NotificationChannel)
+
+    # renderer class - as a string - that will handle the rendering of
+    # this type
+    renderer = StringField()
```
cpennington: Isn't the `renderer` a property of the channel, not of the notification type? The format of the output is going to depend heavily on how the message is getting to the user, right?

edx_notifications/base_data.py
```diff
@@ -0,0 +1,410 @@
...
+            """
+            JSON serializer for objects not serializable by default json code.
+            For now, this means datetime objects.
+            """
+
+            if isinstance(obj, datetime):
+                serial = obj.isoformat()
+                return serial
+
+            return obj
```
cpennington: This doesn't behave properly: if it can't serialize the object, it should raise `TypeError` (the same way that the default argument to `default` does https://docs.python.org/2/library/json.html#json.dump).

edx_notifications/lib/consumer.py
cpennington: You should make sure you call `contracts.disable_all()` when this runs as an independent process, so that there's no overhead to the contracts calls.

edx_notifications/stores/sql/store_provider.py
```diff
@@ -0,0 +1,282 @@
...
+                - unread: Whether to return unread notifications (default True)
+                - type_name: which type to return
+
+        RETURNS: integer
+        """
+
+        return self._get_prepaged_notifications(
+            user_id,
+            filters=filters,
+        ).count()
```
cpennington: It would be good to verify this in a test that hits the paging limit.

edx_notifications/channels/tests/test_channel.py
```diff
@@ -0,0 +1,205 @@
...
+        self.test_msg_type = NotificationType(
+            name='edx_notifications.channels.tests.test_channel.channeltests.foo',
+            renderer='foo.renderer',
+        )
+
+    def cleanUp(self):  # pylint: disable=invalid-name
+        """
+        Test tear down
+        """
+        reset_notification_channels()
```
cpennington: Rather than a `cleanUp` method, you should be calling `self.addCleanup(reset_notification_channels)` in `setUp`. The equivalent method is `tearDown`, but using `addCleanup` is better (it has better guarantees around actually calling the cleanup method.

edx_notifications/data.py
chrisndodge: Well, different NotificationTypes will have different payload schema's, for example, a notification regarding a forum posting will need to present a message regarding the Topic and who created it, etc.. That is why I'm binding a Renderer to a NotificationType. Multiple NotificationTypes can point to the same renderer, for re-use. 

In order to account for different channels needing different renderings, there's a "format" property that is passed into the renderer. You can see an example in edx_notifications/renderers/base.py which is a simple Subject/Body renderer which - for now - only allows for format="underscore" (for client-side rendering).

Let me see if I can fill in some detail and provide more sample code for you.

edx_notifications/channels/tests/test_channel.py
chrisndodge: done. thx

edx_notifications/data.py
cpennington: So each notification type has particular data that needs to be made into a subject/body format, right? In that case, why can't we just have a standard templating method, where the type provides a template (statically), and then the data that populates that template (per user/message pair)?

edx_notifications/stores/sql/store_provider.py
chrisndodge: done. thx.

edx_notifications/server/web/static/edx_notifications/js/vendor/dev/backbone.js
chrisndodge: I moved those out of edx_notifications/server/web and into the testserver. The assumption will be that the hosting application (like the LMS) will provide these files (but we need it in our testserver).

edx_notifications/lib/consumer.py
chrisndodge: Would something like that be in a wsgi.py file? Or some other "on process startup" point? I presume that'd be a globe disabling, right?

edx_notifications/base_data.py
cpennington: I was thinking that `json.dumps(some_datetime) == {'_type': 'datetime', '_value': ISO_FORMAT}`, or something like that. Not sure if that's really better than just guessing based on the value, though.

edx_notifications/base_data.py
cpennington: It might be good to document that, then.

edx_notifications/server/web/static/edx_notifications/js/views/counter_icon_view.js
chrisndodge: Jasmine test harnessing (and initial sample test) can be found at: https://github.com/edx/edx-notifications/pull/15

We're waiting to merge that in after we merge in the RequireJS changes (https://github.com/edx/edx-notifications/pull/16), as we'll have to update the Jasmine harnessing to account for RequireJS

edx_notifications/lib/consumer.py
cpennington: Yeah, that would be in the `wsgi.py` file. Although you can also disable them using an environment variable `DISABLE_CONTRACTS`.

edx_notifications/lib/consumer.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+@contract(user_id='int,>0')
+def get_notifications_count_for_user(user_id, filters=None):
+    """
+    Returns the number of notifications this user has
+
+    ARGS:
+        user_id: The id of the user to query
+        filters: A dict containing the following optional keys
```
cpennington: Would it be better to just have these as optional arguments to this function? Or to have a Filter class that defines the set of valid filters?

edx_notifications/channels/durable.py
```diff
@@ -0,0 +1,51 @@
...
+
+        store = notification_store()
+
+        _msg = store.save_notification_message(msg)
+
+        # create a new UserNotification and point to the new message
+        # this new mapping will have the message in an unread state
+        # NOTE: We need to set this up after msg is saved otherwise
+        # we won't have it's primary key (id)
+        user_map = UserNotification(
```
cpennington: There are still a lot of references to `user_map`, which may be confusing now that the model name has changed.

edx_notifications/channels/durable.py
```diff
@@ -0,0 +1,51 @@
...
+            user_id=user_id,
+            msg=_msg
+        )
+
+        _user_map = store.save_user_notification(user_map)
```
cpennington: Can you get a NotificationUserMap from anything but an existing store? Does it make sense to save a map from one store to another? If not, maybe it would be better for the NotificationUserMap objects provided by the store to have the store that they will save to baked into them, so that you would just call user_map.save().

On the plus side, this might also mean that you could just return the django-backed user_map object directly, since it already implements that interface.

cpennington: Oops. I see that you answered this question in another thread.

edx_notifications/base_data.py
```diff
@@ -0,0 +1,412 @@
...
+    def __set__(self, data_object, value):
+        """
+        Descriptor setter. Be sure to enforce type on setting. But None is allowed.
+        """
+
+        self._assert_has_name()
+
+        value_type = type(value)
+
+        if value and value_type not in self._expected_types:
```
cpennington: This should use an `isinstance` test, rather than `type(value)`.

cpennington:
I'm :+1:  on a raw code level at this point, I think (although I'm coming to agree with @mulby's assertion that Github is bad at reviews of this length).

I have outstanding questions/concerns on the following:
1) Would it be better to expose the underlying storage object mapper, rather than having two level of object mapper?
2) Would it be better to have notification types/channels as code objects, rather than storing them in the database?
3) Should there just be a standard templating engine? Does that really need to be pluggable?

andy-armstrong:
I'm nervous about the RequireJS usage and how it will play with LMS and Studio. This is something we need to resolve for XBlocks too but it hasn't been high on my priority list so far. Let's discuss next week.

chrisndodge:
@karadhya here is the "ongoing review" branch which we have open on our repository, in case you want to see how this is all laying out in code.

edx_notifications/channels/durable.py
```diff
@@ -0,0 +1,90 @@
...
+            )
+            cnt = cnt + 1
+            total = total + 1
+            if cnt == const.MAX_BULK_USER_NOTIFICATION_SIZE:
+                store.bulk_create_user_notification(user_msgs)
+                user_msgs = []
+                cnt = 0
+
+        if user_msgs:
+            store.bulk_create_user_notification(user_msgs)
```
cpennington: Nit: this code might be clearer if you separated the grouping into a separate method (using something like https://stackoverflow.com/questions/312443/how-do-you-split-a-list-into-evenly-sized-chunks-in-python).

edx_notifications/lib/tests/test_lib.py
```diff
@@ -0,0 +1,299 @@
...
+            }
+        )
+
+        def _user_id_generator():
+            for user_id in range(1, 100):
+                yield user_id
+
+        # now send to more than our internal chunking size
+        bulk_publish_notification_to_users(
+            _user_id_generator(),
```
cpennington: Nit: you could just replace `_user_id_generator()` with `xrange(1, 100)`

edx_notifications/stores/sql/store_provider.py
```diff
@@ -0,0 +1,324 @@
...
+        NOTE: This method cannot update existing UserNotifications, only create them.
+        NOTE: It is assumed that user_msgs is already chunked in an appropriate size.
+        """
+
+        if len(user_msgs) > const.MAX_BULK_USER_NOTIFICATION_SIZE:
+            msg = (
+                'You have passed in a user_msgs list of size {length} but the size '
+                'limit is {max}.'.format(length=len(user_msgs), max=const.MAX_BULK_USER_NOTIFICATION_SIZE)
+            )
+            raise BulkOperationTooLarge(msg)
```
cpennington: This behaves differently than

edx_notifications/lib/publisher.py
```diff
@@ -0,0 +1,150 @@
...
+        num_sent = bulk_publish_notification_to_users(
+            User.objects.values_list('id', flat=True).all(),
+            msg
+        )
+
+    """
+
+    if (not isinstance(user_ids, list) and not
+            isinstance(user_ids, types.GeneratorType) and not
+            isinstance(user_ids, ValuesListQuerySet)):
```
cpennington: This is a good candidate for adding to the `@contract` declaration.

    @contract(user_ids="list | GeneratorType | ValuesListQuerySet")

You'll probably need to use `new_contract` to enable `GeneratorType` and `ValuesListQuerySet` as valid contract types:

    new_contract('GeneratorType', types.GeneratorType)
    new_contract('ValuesListQuerySet', ValuesListQuerySet)

cpennington:
I've reviewed up through 39dcfc4, and only had some minor additional comments. I would like to talk over the design of the Renderers system, though. I'm a bit unclear on what usecases it's trying to solve by being pluggable.

chrisndodge:
@cpennington thanks for the additional comments. I'm going to do a quick spike during the Hackathon with regards to defining multiple renderers for each NotificationType.

cpennington:
I've reviewed through `404cbe4`, and have no additional comments.

At what point are you guys going to move onto the per-week or per-sprint "feature" branch, rather than just appending everything to master?

chrisndodge:
@cpennington we're pretty much doing all new stuff on branches now. You can see them in the PR tab.

edx-webhook:
Thanks for the pull request, @chrisndodge! I've created [OSPR-2130](https://openedx.atlassian.net/browse/OSPR-2130) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

Feel free to add as much of the following information to the ticket:
- supporting documentation
- edx-code email threads
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

PR #13, opened 2015-02-04
Added pylru to the dependencies in setup.py

afzaledx:
@chrisndodge, pylru was not being installed, causing tests failures. Just added this  and now the tests are running fine locally.


chrisndodge:
Thanks for catching that. I guess since I'm not pip installing this repo, I hadn't noticed this.

PR #14, opened 2015-02-09
Kashif/bok choy setup

kashifch:
@chrisndodge ,

Please review. I have added the run_bokchoy_tests.sh file which will start the server and run the bok-choy tests.

Added the software requirements in existing test_requirements.txt file.

Added a simple test which just provides dummy credentials on login page and check the error message.

Included the run_bokchoy_tests.sh in travis.yaml script


chrisndodge:
Hi @kashifch thanks for the PR. I'm wondering if we have a timing issue waiting for the testserver to start up. The first time I ran 'bash run_bokchoy_tests.sh', the browser appeared to startup before the testserver has completely started up, so the test was waiting to a page that did not load (because the browser said that it couldn't connect to 127.0.0.1).

Maybe that is why Travis is failing?

chrisndodge:
Tagging @jzoldak to see if someone from TestEng might be of assistance regarding best practices to get bok-choy (as well as Jasmine) to run reliably in Travis. @jzoldak, @kashifch is out test engineer from Arbisoft that is helping us with this project.

kashifch:
Hi @chrisndodge , I will try to find a method to verify that server has started before running the tests. I have tested the .sh file locally but haven't tried it with travis, I will also try to find a way to test the travis locally

run_bokchoy_tests.sh
```diff
@@ -0,0 +1,29 @@
...
+    kill $(ps aux | grep "[m]anage.py" | awk '{print $2}')
+}
+
+echo "Stop any already running servers..."
+stopServers
+
+echo "Starting Notifications Server..."
+./manage.py runserver --settings=testserver.settings --noreload &
+
+
```
chrisndodge: maybe we could use 'curl' here to wait for server to respond before proceeding?!?

chrisndodge:
@kashifch also please add your name to the CONTRIBUTORS file.

chrisndodge:
Also, it seems like the Python unit tests are picking up your bok-choy test. You can see this by running './run_tests'

I'll add the edit.....

kashifch:
@chrisndodge , sorry for botheration but I am having some issues with my git repository, couldn't remove the conflicts right now

I think I found the issue in .travis.yml, it is just an incorrect file name 

I think it will work correctly if you replace
- bash ./run_bok_choy_tests.sh

with 
- bash ./run_bokchoy_tests.sh

chrisndodge:
@kashifch OK, I added a waitstep for the server to respond, but Travis is still failing. I think it'll be best to talk to TestEng about running in Travis environments. If we can't get this resolved within a day or two, let's remove the integration into Travis so we can proceed to merge in the manual kick offs of acceptance tests.

benpatterson:
@kashifch You might want to try setting up a couple of things related to browser usage first. 

I suggest adding the DISPLAY env var, like so:
export DISPLAY=:"99.0" which will set it to go headless.

Also, in a before_script section, you'll want to have:
- "sh -e /etc/init.d/xvfb start"

The script looks very similar to the dashboard repo, and you can see we set up DISPLAY and xvfb there...in the .travis.yml. https://github.com/edx/edx-analytics-dashboard/blob/master/.travis.yml#L8-L9

chrisndodge:
@kashifch I applied Ben's suggestions. It seems better, but now there is a failing assertion in the bok-choy test itself as opposed to a "browser closed unexpectedly" error. Can I pass this back to you?

chrisndodge:
Ah, I think I know why it is failing, it has not created the test database:

We'll need to add a:

./manage.py syncdb --settings=testserver.settings
./manage.py migrate --settings=testserver.settings

I'll give that a shot....

chrisndodge:
OK, I think we got the bok-choy to work. I think the failure is due to the decreased code coverage.

testserver/test/acceptance/pages/login_page.py
```diff
@@ -0,0 +1,26 @@
+__author__ = 'Kashif'
```
chrisndodge: I'm not sure we typically use these __author__ tags in Open edX code base. Normally we put our names in the CONTRIBUTORS file.

testserver/test/acceptance/test_login.py
```diff
@@ -0,0 +1,26 @@
+__author__ = 'kashif'
```
chrisndodge: ditto

run_bokchoy_tests.sh
chrisndodge: I added it (see below)

testserver/test/acceptance/pages/login_page.py
```diff
@@ -0,0 +1,26 @@
...
+from bok_choy.page_object import PageObject
+from . import base_url
+
+
+class LoginPage(PageObject):
+
+    url = base_url
+
+    def is_browser_on_page(self):
+        #check the browser title
```
chrisndodge: Although we don't run pylint/pep8 on these files, it'd be good to follow standard conventions, like docstrings on every method:

def foo(self):
    """
    comment here
    """

run_bokchoy_tests.sh
```diff
@@ -0,0 +1,36 @@
...
+echo "Waiting for testserver to fully start up..."
+
+until $(curl --output /dev/null --silent --head --fail http://127.0.0.1:8000); do
+    printf '.'
+    sleep 2
+done
+
+echo "Running acceptance tests..."
+python testserver/test/acceptance/test_login.py
```
chrisndodge: what should we do here when we have a whole test suite, is there a way to trigger the execution of all of them? Do we have to do some bash shell magic to do a 'find...' (all test_*.py) and iterate over the list?

kashifch:
Hi @chrisndodge ,

Done following changes

1) Added my name in authors file and removed the name from bok_choy files
2) used docstrings for commenting
3) added another test
4) added a loop in shell script to run all the test files in acceptance test folder instead of running individual tests

benpatterson:
@karadhya @chrisndodge In terms of making sure you run all the tests in the acceptance folder, running a loop can get you there, but as you add more tests, it might not be as straight-forward to count the pass/fails/etc. 

I suggest using nose (`nosetests`). You can just feed nosetests a directory and it will find all test files (whose names start with `test`) and run those.

Aside from the advantage of counting your passes and fails, it is also our standard for test execution for various other reasons. So, could you do that instead?

kashifch:
@benpatterson , thanks for the suggestion,

chrisndodge:
@benpatterson thanks for the feedback. Do you mind a +1 here? We're trying to do "ongoing reviews" during the project where 2nd reviewers basically check-off that "yeah, this looks kinda right" on a ongoing basis so that we avoid big surprises when we're doing final reviews on the project (in about 6 weeks).

benpatterson:
Yep, LGTM guys. :+1:

PR #16, opened 2015-02-13
Cdodge/migrate to require js

chrisndodge:


chrisndodge:
@kashifch @afzaledx @muhhshoaib I converted over the Notification BackboneJS to use the RequireJS framework. Hopefully this will be helpful for writing Jasmine tests as all the dependencies are now explicitly declared in all of the modules. Take a quick look and see if it makes sense, then I'll merge it into master.

FYI @andy-armstrong, I hope this is the right approach. I noticed a RequireJS plugin (https://www.safaribooksonline.com/library/view/developing-backbonejs-applications/9781449328535/ch04s03.html) to allow for the declaring of Underscore Template dependencies, so I'm using that for the async loading of the statically declared Underscore templates. From manual testing, it seems to do what I want it to do.

I hope that (the declaring of underscore templates in RequireJS) plays nicely with Jasmine tests. This is all very new to me, so any guidance would be very helpful.

Then I'll take a look at that JS Optimizer thingy at some point next, now that we have a top level entry point into the JS with most dependencies statically declared.

andy-armstrong:
@chrisndodge the library you are using (requirejs.text) is the library we HipChatted about using last week, and is the way I am recommending that we lazily load templates. It is somewhat documented on my wiki page (we have only used it in hackathon projects so far):

https://openedx.atlassian.net/wiki/display/TNL/How+to+add+a+new+page+to+LMS+or+Studio

Have you tried to get this to work in the LMS or Studio yet? I'm not sure how this JavaScript will interact with the page's own RequireJS configuration. @cpennington, I think that there's a similar challenge here with using RequireJS in xblocks. I think we should all chat about this next week and determine the best approach to use.

BTW, here's my page where I've been collecting information about how to use RequireJS in the LMS:

https://openedx.atlassian.net/wiki/display/TNL/RequireJS+in+the+LMS

andy-armstrong:
@chrisndodge I'm particularly nervous about how this is going to interact with pages that are themselves using RequireJS Optimizer. This is not being used yet in the LMS, but most Studio pages are using it. Do you know yet which pages are going to show the notification widget?

chrisndodge:
closing this PR as we have a different branch where @kashifch worked out the Jasmine/RequireJS stuff at: https://github.com/edx/edx-notifications/pull/17

PR #17, opened 2015-02-17
Kashif/jasmine with requirejs

kashifch:
@chrisndodge ,

I have completed the integration of jasmine with require js.

Just one issue.

In your js files you have used paths like

screen shot 2015-02-17 at 8 37 45 pm

This was causing my tests to fail as tests were able to load first js file but not the files referenced in it. I have changed al the paths to variable names which are stored in centralised locations and after that the test started running successfully

screen shot 2015-02-17 at 8 40 13 pm


chrisndodge:
Hi @kashifch,

Thanks. I'll take a look at this PR. Also there's a merge conflict that I will look into.

I'll also talk with @andy-armstrong to resolve the open questions about the use of RequireJS and how that might integrate into the LMS and Studio.

chrisndodge:
@kashifch builds are now passing, must have been a transient problem. I'll follow up with @andy-armstrong about high-level feedback on the RequireJS approach.

chrisndodge:
@kashifch just a reminder to please add Jasmine testing instructions (including the NodeJS and dependencies installation). Thanks.

chrisndodge:
@kashifch I made a commit here to utilize the RequireJS namespacing that the LMS uses. I figured - to help out integrating into the LMS - we should at least use that same pattern.

chrisndodge:
@kashifch @andy-armstrong I'm going to merge this in as failure to do so will block some sprint stories which need to start tomorrow. I'll take it on good faith that we'll figure out a way to work out any kinks regarding integrating this use of RequireJS in the Open edX LMS/Studio when the time comes. If not, then we should be able to go back to non-requireJS.

PR #20, opened 2015-02-24
creating separate test db for bok-choy

kashifch:
added a separate bokchoy_settings file which will create test_notifications db, the shell script will now use this setting file to create db and after the tests are complete the db is deleted using simple rm command

### DIFF
--- a/run_bokchoy_tests.sh
+++ b/run_bokchoy_tests.sh
@@ -1,4 +1,5 @@
-i#!/usr/bin/env bash
+#!/bin/bash bash
+
 
 # this stops the django servers
 stopServers() {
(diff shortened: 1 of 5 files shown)


chrisndodge:
thanks @kashifch seems to work. Also as a side note, I got a sporadic error regarding verifying the content of a notification. My guess is that it is a timing issue where the HTML isn't fully rendered before you are doing the content checking. These timing issues have been a problem for a number of people and there might be some best practices with bok choy to make sure that checks aren't don't before data is available. Is there anyone at Arbisoft that works with @jzoldak that might be able to give you some guidance?

PR #21, opened 2015-02-24
as a stop gap measure to fix bok choy test timing issues, put a sleep st...

chrisndodge:
...atement

### DIFF
--- a/testserver/test/acceptance/pages/logged_in_home_page.py
+++ b/testserver/test/acceptance/pages/logged_in_home_page.py
@@ -1,6 +1,7 @@
 from bok_choy.page_object import PageObject
 from . import user_name
 
+from time import sleep
 
 class LoggedInHomePage(PageObject):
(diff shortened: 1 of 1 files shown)


jzoldak:
This makes me sad. :(
And I fear that it's probably not going to solve the underlying flakiness.
At the very least it should be tagged with a TODO and put on the backlog.

kashifch:
@jzoldak , 
I have used the wait for element check 
self.wait_for_element_visibility('.edx-notifications-list', 'Notification messages list not found')
return self.q(css='.edx-notifications-list').text

I think the problem is that it finds the element but the text of the element is not loaded completely, so it returned an empty list. Would it help if I use wait_for_ajax here?

chrisndodge:
Sorry to cause scurrying, I would have put another message on the commit saying this was going to be fixed by Kashif.

PR #22, opened 2015-02-25
used wait_for_ajax instead of time.sleep

kashifch:
@chrisndodge ,  I have added the wait for ajax command instead of time.sleep. I have tested it on local machine and also on network, so far I haven't seen any failure. Let's see if this fixes the issue


chrisndodge:
Thanks @kashifch for researching this. Looks likes it is passing. Merging.

PR #23, opened 2015-02-26
(WIP) SOL-346 Notification List pane implemented in the edx-notifications

muhhshoaib:


edx_notifications/server/api/consumer.py
```diff
@@ -211,6 +212,25 @@ def post(self, request, msg_id):
         return Response({}, status.HTTP_200_OK)
 
 
+class MarkNotifications(AuthenticatedAPIView):
```
chrisndodge: As this is the first POST method on this API, we'll have to investigate how to ingrate CSRF tokens to prevent cross-site scripting. I'll do some researching.

edx_notifications/server/api/consumer.py
```diff
@@ -211,6 +212,25 @@ def post(self, request, msg_id):
...
+    """
+
+    def post(self, request):
+        """
+        HTTP POST Handler which is used for such use-cases as 'mark as read'
+        """
+        try:
+            # this will raise an ItemNotFoundError if the user_id/msg_id combo
+            # cannot be found
+            mark_all_user_notification_as_read(request.user.id)
```
chrisndodge: We should probably allow for a POST variable to say whether all UserNotifications should be marked as read OR unread. This isn't the current use-case, but I'd anticipate we might want to support this in the future.

edx_notifications/server/web/static/edx_notifications/js/views/notification_pane_view.js
```diff
@@ -124,7 +134,60 @@ define([
...
+            /* make the async call to the backend REST API */
+            /* after it loads, the listenTo event will file and */
+            /* will call into the rendering */
+            var self = this;
+            self.$el.addClass('ui-loading');
+            this.collection.fetch(
+                {
+                    data:{
+                        mark_as: 'read',
+                        csrfmiddlewaretoken: $('input[name="csrfmiddlewaretoken"]').prop('value') }
```
chrisndodge: Ah, I see CSRF tokens here. Thx!

edx_notifications/server/api/consumer.py
afzaledx: @chrisndodge, Shoaib did add the flag to mark all as unread/read, but I asked him to remove it since I cannot think of any situation where I would want to mark _all_ my notifications as unread. Marking a selected notification as unread does make sense though.

edx_notifications/server/web/static/edx_notifications/templates/notification_pane.html
```diff
@@ -1,6 +1,35 @@
...
+            <% if (typeof user_notifications == 'undefined' || user_notifications.length == 0) { %>
+                <li>
+                    <p class="description">You have no unread notifications.</p>
+                </li>
+            <% } else { %>
+            <% _.each(user_notifications, function(user_notification){ %>
+
+            <li>
+                <h3 class="borderB padB5 uppercase bold marB10">Annoucments</h3>
+                <h3 class="marT0"><a href="">Welcome to McKinsey Academy</a></h3>
```
chrisndodge: We'll have to make that McKinsey Academy some template variable that is passed in. I'll have to think how to do that...

chrisndodge:
Seems like we need to add some unit test coverage to some of the API changes to get Coveralls to not fail (it won't allow decrease in test coverage).

edx_notifications/server/web/static/edx_notifications/templates/notification_pane.html
```diff
@@ -1,6 +1,35 @@
...
     <% _.each(user_notifications, function(user_notification){ %>
         <li class='edx-notifications-list-item'><%= user_notification.html %></li>
     <% }); %>
 </ul>
+<div class="container">
+    <div class="content">
+        <h2>Notifications</h2>
+        <div class="actions">
+            <ul class="notifications_list">
+                <li class="active"><a id="unread_notifications" href="#">View unread</a></li>
```
chrisndodge: We'll have to research how to support localization in Underscore/Backbone. I'm not sure, so I'll try to ask someone here. @andy-armstrong perhaps?

afzaledx:
We quickly pushed partial work just so you could have a look at the code changes and see it running. It's not finished yet.

chrisndodge:
Gotcha, sorry. I'll save detailed comments for later.

chrisndodge:
@muhhshoaib @afzaledx just ran the branch now. Very impressive in such a short amount of time. I have some comments, but I don't know if you want the feedback now or wait until you finish up.

afzaledx:
Thanks. really appreciate that. It was really enjoyable looking at everything coming together. We do plan to make changes but please feel free to give any feedback.

chrisndodge:
@muhhshoaib @afzaledx @asadiqbal08 Here are some user-perspective notes (I'll look at code later). I'm just jotting things down as I see them, some of which you might already be aware of:

1) As the user clicks the various tabs (View unread/ View all), the 'selected' style should be updated to reflect which tab is being shown

2) When 'Mark as read' is clicked, the Notification Icon should also be refreshed to show that there are 0 unread notifications

3) In the pane, the date of the Notification should be formatted in some readable format, not ISO 8601 format that the API returns. Look at a 'datejs' utility which is good for parsing ISO8601 and reformatting. Use the spec as shown in the PDF document for the feature as a guide to what formatting to use.

4) I will have to come up with a way to get the header/sub header display names from the API. Right now it appears to be hard coded to say 'Announcements" and "Welcome to <app_name>".

5) In the "view unread" tab, all notifications of the same type should be grouped together under the same header. There should not be a new header between notifications of the same type. You will need to sort the collection by type name before passing it to the Underscore template.

6) While we don't need to present this in other languages, we should try to make the Underscore templates localizable. I don't know how to do this offhand. I'll try to research over the weekend.

7) The CSS class names that we are using are a bit generic and I'm concerned there might be style collisions when we integrate this into the application UI. For example, in notification_pane.html, we have:

class="container"
class="content"
:
:

We should change this to something more specific like,

class="edx-notifications-container"
class="edx-notifications-content"
:
: 
etc.

Thanks, keep up the good work!

edx_notifications/server/web/static/edx_notifications/templates/notification_pane.html
```diff
@@ -1,6 +1,35 @@
...
     <% _.each(user_notifications, function(user_notification){ %>
         <li class='edx-notifications-list-item'><%= user_notification.html %></li>
     <% }); %>
 </ul>
+<div class="edx-notifications-container">
+    <div class="edx-notifications-content">
+        <h2>Notifications</h2>
+        <div class="actions">
+            <ul class="notifications_list_tab">
+                <li id="unread_notifications" class="active"><a  href="#">View unread</a></li>
```
chrisndodge: can you use class names rather than id's? You'll have to update the jQuery selectors as well.

edx_notifications/lib/tests/test_lib.py
```diff
@@ -208,6 +209,47 @@ def test_bulk_publish_notification_orm_query(self):
...
+        for i in range(10):
+            msg = NotificationMessage(
+                namespace='test-runner',
+                msg_type=self.msg_type,
+                payload={
+                    'foo': 'bar'
+                }
+            )
+            publish_notification_to_user(self.test_user_id, msg)
+
```
cpennington: It might be worth an assertion here about the number of unread messages, just to make sure this test isn't vacuous.

edx_notifications/server/api/consumer.py
```diff
@@ -117,7 +118,7 @@ def get(self, request):
             return Response({}, status.HTTP_400_BAD_REQUEST)
 
         cnt = get_notifications_count_for_user(
-            request.user.id,
+            int(request.user.id),
```
cpennington: When is `user.id` not an `int`?

chrisndodge: Once I got this running in a devstack (as opposed to localdev running sqlite) it appears that request.user.id is a long.

I didn't see a way to have a @contract() which allowed for multiple types. Do you have an idea?

cpennington: Yes, you can have a contract like "int | long"

chrisndodge: Sorry, no dice. My unit tests are now saying:
ERROR: Failure: ContractSyntaxError (Unknown identifier 'long'. Did you mean 'None'? (at char 4), (line:1, col:5)

 line  1 >int|long,>0
              ^
              |
              here or nearby)

'long' doesn't seem to be a contract keyword...

chrisndodge:
Closing PR to re-open in order to merge into an intermediate branch.

PR #24, opened 2015-02-26
Add ability to pass in global_variables, such as the app_name, into the ...

chrisndodge:
...Backbone views

### DIFF
--- a/edx_notifications/server/web/utils.py
+++ b/edx_notifications/server/web/utils.py
@@ -18,7 +18,7 @@ def get_template_path(template_name):
     )
 
 
-def get_notifications_widget_context():
+def get_notifications_widget_context(override_context=None):
     """
     As a convenience method, this will return all required
     context properties that the notifications_widget needs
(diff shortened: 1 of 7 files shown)


chrisndodge:
@afzaledx @muhhshoaib I created a branch off of your branch to be able for the application host container to pass along global variables, such as 'app_name' all the way through the Backbone views. Can you look at this and I'll merge it into your working branch.

edx_notifications/server/web/utils.py
```diff
@@ -40,10 +40,16 @@ def get_notifications_widget_context():
...
+            'app_name': 'Your App Name Here',
+        },
         'view_templates': {
             'notification_icon': get_template_path('notification_icon.html'),
             'notification_pane': get_template_path('notification_pane.html'),
         }
     }
 
+    if override_context:
+        context.update(override_context)
```
muhhshoaib: @chrisndodge what is override_context here. What i am getting is that some extra settings that may be needed in the future ?

muhhshoaib:
looks good :+1:

edx_notifications/server/web/utils.py
chrisndodge: This is where an application can pass in additional context, for example, the app_name which only the application tier knows about.

PR #31, opened 2015-03-02
Cdodge/poll

chrisndodge:


cpennington:
Do yet have a sense of the operational impact of this polling?

chrisndodge:
@cpennington I've written comments in the source code on the use of short-polling only for small installations and demo purposes. We're planning to write a long-poll version as well which will be backed by a Comet server (like nginx's plugin) which will have better concurrency/scaling characteristics.

asadiqbal08:
@chrisndodge that mp3 tone is pushed mistakenly, Yeah that tone is funny but i used another mp3 file.

PR #44, opened 2015-03-04
Cdodge/fix pane tabs

chrisndodge:


chrisndodge:
@muhhshoaib here's the fixes. Plus looking at the code, I realized that we could simply use Underscore's groupBy() and orderBy() functionality. So I did a bunch of rewriting of the JS and tightened up some of the management of the tab highlighting.

Also, I reverted your change to sort the SQL resultset on the server. The data from the StoreProvider should not be assumed by any clients to be ordered according to whether we want unread or all. The fact that we are grouping together by Type at the presentation tier, may not be valid in other use cases or presentation tiers.

Hope this makes sense.

Once I get a passing build I'm going to merge this in because we have a broken mcka-master branch that has to get fixed. But please look at the changes.

PR #47, opened 2015-03-05
changed the title of the unread notifications tab

muhhshoaib:
@chrisndodge i have changed the title of the unread notifications tab.


chrisndodge:
Thanks. This will be a short-term hack, as ultimately, we'll have to pull in group names from the server.

PR #49, opened 2015-03-05
Jasmine tests for Notification Pane View

kashifch:
Hi @chrisndodge ,
Please review.
I have added some jasmine tests for notification pane view.
Introduced line breaks in fixtures for readability


chrisndodge:
@kashifch thanks for keeping to take the initiative!

PR #58, opened 2015-03-06
added the div tag inside the notification underscore template so that mc...

muhhshoaib:
@chrisndodge I have added the div inside the notifications pane template. This need to be in the mcka:master, the css changes that we did in the mcka_repo depends on that div to be there so that they can load perfectly. Basically we have added a fixed div so that the user can scroll the notifications, the notifications tabs don't scroll, they will remain in the same position. only the notifications items will scroll down.

Thanks


chrisndodge:
@muhhshoaib sorry, I have to revert this change. It's causing some rendering problems in McKinsey application. I'd rather have the top region scroll than have these visual errors.

PR #59, opened 2015-03-06
Afzaledx/clickthru

chrisndodge:


edx_notifications/server/web/static/edx_notifications/js/views/notification_pane_view.js
```diff
@@ -257,6 +256,41 @@ var NotificationPaneView = Backbone.View.extend({
             }
         );
     },
+    visitNotification: function(e) {
+        var messageId = $(e.currentTarget).find('span').data('msg-id');
+        var clickLink = $(e.currentTarget).find('span').data('click-link');
+
+        if (this.selected_pane === "unread") {
+            this.collection.url = this.endpoints.user_notification_mark_read.replace("/0", "/" + messageId);
```
chrisndodge: hrm. This seems like we'll need to fix the endpoints that are getting passed into the backbone app. This URL manipulation shouldn't be necessary. I'll look into it.

edx_notifications/server/web/utils.py
```diff
@@ -50,6 +50,9 @@ def get_notifications_widget_context(override_context=None):
             'user_notifications_all': (
                 '{base_url}?read=True&unread=True'
             ). format(base_url=reverse('edx_notifications.consumer.notifications')),
+            'user_notification_mark_read': (
+                '{base_url}'
+            ). format(base_url=reverse('edx_notifications.consumer.notifications.detail', args={'0'})),
```
chrisndodge: I would like to see if we can pass in args={''} so we don't get that trailing 0 that has to get fixed up in JavaScript.

edx_notifications/server/web/static/edx_notifications/js/views/notification_pane_view.js
chrisndodge: Ug, due to the URL regex, it seems like we *have to* pass in an integer :-( here.

Maybe I see a solution: http://stackoverflow.com/questions/14351048/django-optional-url-parameters

chrisndodge:
@afzaledx OK, I think I found a way to not have to remove that trailing '/0' to the URL path. I just added a second URL pattern which is parameterless.

edx_notifications/server/web/static/edx_notifications/templates/renderers/forums/comment_upvoted.html
```diff
@@ -1 +1 @@
-<%= action_username %> has upvoted your comment on a discussion thread titled '<%=thread_title %>'.
+<%= action_username %> has upvoted your comment on a discussion thread titled <a href="javascript: void(0)">'<%=thread_title %>'</a>.
```
chrisndodge: Not sure I understand why this <a> tag is here. Can you explain?

chrisndodge:
Once we get a passing build, I'll merge this in.

One question: I notice on the on-hover over the message, the cursor changes to that "text cursor" rather than a pointer. We like a pointer since it implies that clicking on it will click through to something else rather than a text cursor which implies that HTML text can be selected.

PR #67, opened 2015-03-07
Additions and modifications in bok-choy tests

kashifch:
1) Added the test to verify that clicking on a notification redirects user to a specific page (this test is a bit rough at edges as the functionality is not fully developed, currently all notifications are redirecting user to a new page and as the page is not valid, I have to use back button and then login/logout to record the after effects)

2) Added the tests to verify that clicking on notification removes it from unread notifications tab and also decreases the notification count on main page by 1.

3) Added a login function which summarises all login related statements and used this function in all tests

4) Added and modified some wait checks for slower internet connections

### DIFF
--- a/edx_notifications/server/web/static/edx_notifications/js/test/spec/notification_pane_view_Spec.js
+++ b/edx_notifications/server/web/static/edx_notifications/js/test/spec/notification_pane_view_Spec.js
@@ -8,32 +8,45 @@ describe("NotificationPaneView", function(){
             '<span class="edx-notifications-count-number"></span>' +
             '</div>' +
             '<script type="text/template" id="notification-pane-template">' +
-            '<div class="edx-notification-pane">' +
-            '<div class="edx-notifications-container">' +
-                    '<div class="edx-notifications-content">' +
-                        '<h2>Notifications</h2>' +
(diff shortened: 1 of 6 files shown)


chrisndodge:
@kashifch thanks! Sorry to see you working on a weekend! :-(

Thanks in any case.

chrisndodge:
To help with tests, I can simply make a valid redirect target. If you are not working on the branch, I can add that and pass it back to you.

kashifch:
Nah, I think it is okay for now, by valid url I meant a target page belonging to course from where I could always return to the home page by clicking on the course name. Any other url and I would have to use the back button or url to get back to homepage, anyway the tests are running even on a slow internet so let's keep it as is until it is merged with a real course(though if you need to change the url for any other reason, do go ahead, it shouldn't break the tests)

kashifch:
Today
Added some bok-choy tests to test notifications which does not redirect user to another page
updated the fixture of notification pane for in jasmine test
changed bok-choy and jasmine tests to accommodate the exclusion of hide_pane
added jasmine test for visitNotification function calling

chrisndodge:
OK, I'll let this build. Since it's just tests I'll merge it in afterwards.

PR #72, opened 2015-03-09
chagned the notification structure and removed the hide link fron the notification pane

muhhshoaib:


chrisndodge:
Thanks. I'd prefer this being an configuration option that can be passed into the Backbone app, so sites can determine if they want a 'hide' link or not.

chrisndodge:
I had to resolve some merge conflicts....

kashifch:
changed some tests to include handling of hide link settings, The tests now read the settings from bokchoy_settings file and run accordingly, The test to check hide link is skipped if the variable is set to false in setting file, also the list for tab verification is created on the basis of this variable

PR #74, opened 2015-03-10
Provide for arbitrary groupings of presenting notifications plus do some...

chrisndodge:
... McKA confirming to the PDF


chrisndodge:
@afzaledx @muhhshoaib FYI, this isn't going to merge until after the demo.

kashifch:
hmm, the test fails due to change in sort order, I was picking the notification on the top, I think I would need to look for groups groups in the test

chrisndodge:
Seems like the new groups/ordering is breaking a bok choy test because it expects some particular content to be at the top. I'll fix the test tomorrow.

kashifch:
sorry for the intrusion but since I figured out the solution I am taking the liberty to update the tests. I will upadate my branch with same changes as well to avoid any conflicts

PR #80, opened 2015-03-14
efficiency and usability enhancements

kashifch:
@chrisndodge ,

Here are the changes
1) All the wait and empty promises default time has been decreased to 20 sec
2) User can now provide the test name after the bash command to run a specific test, I have updated the readme and also shortened the test names
3) Some loops have been cut down to save time
4) All logs will be placed in Logs folder inside acceptance directory (script will create the folder)


kashifch:
One observation, it seems that hide link is not working, the reason bokchoy tests passed was that by default the hide link functionality is turned off

kashifch:
Also you can delete or close the kashif/bokchoy_tests_efficiency branch as all it's changes have been included the current branch

run_bokchoy_tests.sh
```diff
@@ -19,7 +28,7 @@ fi
 echo "Creating new test db and sync data..."
 ./manage.py syncdb --noinput --settings=testserver.bokchoy_settings
 
-echo "Migrate data"
+ech "Migrate data"
```
chrisndodge: is this a typo? shouldn't it be echo?

chrisndodge:
Thanks. Just one general suggestion as I made before. Can we make the timeout defined in one place, e.g.

TEST_TIMEOUT = 20

and then have all the tests use that definition. That way we can change the timeout value in just one spot and have all the tests reflect that change.

kashifch:
1) typo correction in shell script
2) defined default_timeout as global variable in __init__ file, now all waits and promises are using this variable

PR #84, opened 2015-03-16
Kashif/minor cahnges in tests

kashifch:
Added engagement leaderboard notification in bok-choy
added some jasmine tests

### DIFF
--- a/edx_notifications/server/web/static/edx_notifications/js/test/spec/counter_icon_view_Spec.js
+++ b/edx_notifications/server/web/static/edx_notifications/js/test/spec/counter_icon_view_Spec.js
@@ -2,6 +2,7 @@ describe("CounterIconView", function(){
 
     beforeEach(function(){
         this.server = sinon.fakeServer.create();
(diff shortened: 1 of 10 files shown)


kashifch:
added support for PhantomJS2 as tests which included prototypes or bindings were failing on PhantomJS older version

kashifch:
Hi @chrisndodge ,
In the last commit I have changed something from notification_pane_view.js file, the .xns-hide_pane was written as .xns-hide_pane in it due to which the click event on hide button was not working, As by default we have the hide link settings turned to false so bok-choy test was skipped, I have turned the setting as true in bok-choy test settings so the test will now run every time. Overall it seems to be working fine but do take a look to verify that this change does not break anything else

kashifch:
seems like we have another flaky test, will look into it tonight

kashifch:
added a couple of waits for flakiness, let's see if these work

chrisndodge:
@kashifch thanks for catching the hide-pane bad class name.

PR #86, opened 2015-03-18
Asadiqbal08/sol 512

asadiqbal08:
@chrisndodge kindly review the change set. 

Changes are : 

- Added a sibling div element (class = "xns-close-item") to "xns-item-body"
- 'x' character  will appear on hover and only under the 'unread' pane/tab.
-  clicking on 'x' does not cause the user to lose scroll position.

1) notification has 'click_link' url: 
  - when a user will click on character 'x' , notification will be marked as read and removed from the pane. 
  - when a user will click on notification area instead of 'x' then notification will marked as read and redirected to the click_link url. (default behavior)

2) notification has NOT a 'click_link' URL : 
  - if user click on 'x' or anywhere on the notification then it will follow the same default behavior.

let me know if anything else is required in it. 

Thanks.

### DIFF
--- a/edx_notifications/server/web/static/edx_notifications/js/views/notification_pane_view.js
+++ b/edx_notifications/server/web/static/edx_notifications/js/views/notification_pane_view.js
@@ -484,7 +484,7 @@ var NotificationPaneView = Backbone.View.extend({
                         if (clickLink && !$(e.target).hasClass("xns-close-item")) {
                             window.location.href = clickLink;
                         }
-                        else if ($(e.target).hasClass("xns-close-item")) {
+                        else if ($(e.target).hasClass("xns-close-item") && $(".xns-items li").length > 1) {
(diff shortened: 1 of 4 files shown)


chrisndodge:
Thanks @asadiqbal08, I'll take a look shortly.

chrisndodge:
@asadiqbal08 head's up. I just made one small change. I added a <span> around the 'x' character so that I can apply specific style rules to it. Once I get a completed and passing build, I'll merge it in.

Thanks & good job.

chrisndodge:
@asadiqbal08 OK, adding the span is causing the Backbone view to properly handle the click events. I did a little refactoring, making a new 'closeNotification' handler specifically attached to the xns-close-item class, so I can specifically target that element. There's a little bit of duplicate code, that we can refactor to use a combined shared utility function later, but I'd like to merge this so I can deploy to the test server.

PR #91, opened 2015-03-24
SOL-552 cross browser compatibility issue.

muhhshoaib:
@chrisndodge here is the PR for the IE browser issues. I hope this will fix this.

I had some problem with the date.js library in the IE9. I skipped those issues and worked on the browser issues. The integration server is working fine with dates and i think this is only in my local server.

### DIFF
--- a/edx_notifications/server/web/templates/django/notifications_widget.html
+++ b/edx_notifications/server/web/templates/django/notifications_widget.html
@@ -77,6 +77,7 @@ page that is using this widget
 </script>
 
 <script type="text/javascript">
+    Backbone.$.ajaxSetup({ cache: false });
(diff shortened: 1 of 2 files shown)


asadiqbal08:
@chrisndodge i was looking that .bind() function is not working in IE 9. i was using it in binding of autoRefreshNotifications in counter_icon_view.js file. Kindly let me know is there any story that testing the functionality of autoRefreshNotifications in IE 9 or 10 ?

chrisndodge:
@asadiqbal08 if you are seeing problems with auto-refresh on IE9, let's treat that as a separate bug. If you believe there is a defect, I'll write up a JIRA bug ticket.

PR #93, opened 2015-03-25
SOL-564 IE9 not auto refreshing

asadiqbal08:
@chrisndodge kindly review the changes. 
- Fixed: IE9 not auto refreshing

### DIFF
--- a/edx_notifications/server/web/static/edx_notifications/js/views/counter_icon_view.js
+++ b/edx_notifications/server/web/static/edx_notifications/js/views/counter_icon_view.js
@@ -36,7 +36,8 @@ var CounterIconView = Backbone.View.extend({
       /* adding short-poll capabilities to refresh notification counter */
       if(this.refresh_watcher.name == 'short-poll'){
           var period = this.refresh_watcher.args.poll_period_secs;
-          setInterval(this.autoRefreshNotifications.bind(window, this), period * 1000);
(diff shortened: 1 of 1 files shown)


edx_notifications/server/web/static/edx_notifications/js/views/counter_icon_view.js
```diff
@@ -113,15 +114,19 @@ var CounterIconView = Backbone.View.extend({
...
+ autoRefreshNotifications: function(counterView) {
      var currentModel = new CounterIconModel();
      currentModel.url = counterView.model.url;
      currentModel.fetch().done(function(){
          // if notification counter is incremented.
          if(currentModel.get('count') > counterView.model.get('count')){
-             var notification_alert = new Audio(counterView.view_audios.notification_alert);
              counterView.model = currentModel;
-             notification_alert.play();
+
```
chrisndodge: Hi, one comment unrelated to this PR fix, but I just realized that an audio alert is optional, so we should do a test to see if counterView.view_audios.notification_alert is undefined (or null), which should not attempt to play an audio tone.

chrisndodge: I'll merge this for now, so that Kashif can validate this fix, and I'll write up a new JIRA to make the audio optional.

PR #98, opened 2015-03-27
bugs fixes as highlighted by Kashif in edx-notifications

asadiqbal08:
@chrisndodge kindly review some fixes as highlighted by Kashif in edx-notifications.

### DIFF
--- a/edx_notifications/server/web/static/edx_notifications/js/views/counter_icon_view.js
+++ b/edx_notifications/server/web/static/edx_notifications/js/views/counter_icon_view.js
@@ -132,9 +132,8 @@ var CounterIconView = Backbone.View.extend({
                  notification_alert.play();
              }
              counterView.render();
-             if (counterView.notification_pane) {
-                 counterView.notification_pane.collection.url = counterView.options.endpoints.user_notifications_unread_only;
-                 counterView.notification_pane.hydrate();
+             if (counterView.notification_pane && counterView.notification_pane.selected_pane == "unread") {
+                 $('.xns-unread-action').trigger('click');
              }
          }
      });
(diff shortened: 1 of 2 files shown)


edx_notifications/server/web/static/edx_notifications/js/views/counter_icon_view.js
```diff
@@ -132,9 +132,8 @@ var CounterIconView = Backbone.View.extend({
                  notification_alert.play();
              }
              counterView.render();
-             if (counterView.notification_pane) {
-                 counterView.notification_pane.collection.url = counterView.options.endpoints.user_notifications_unread_only;
-                 counterView.notification_pane.hydrate();
+             if (counterView.notification_pane && counterView.notification_pane.selected_pane == "unread") {
```
chrisndodge: I'm not sure I'm following this logic. Can you explain? Also, in general I'd prefer that you directly call into javascript rather than simulating a mouse click. Thanks.

asadiqbal08: @chrisndodge here I am ensuring that selected pane should be 'unread'. 
Because i was seeing that if a user is under 'view all' tab/pane and if a notification(s) arrives at that time then unread notifications are rendering under 'view all' pane. So i am just rendering the unread notifications under the 'unread' pane.

chrisndodge: Ah, now it makes sense to me. Thanks. Any chance we can refactor the code a bit so we don't need to do a 'fake click'?

asadiqbal08: @chrisndodge i have updated the code. kindly look into it.

chrisndodge:
@asadiqbal08 I've reverted this merge. I think I narrowed down to our failing Travis builds to this merge, but I'm not sure. Perhaps there's some edge case happening here?

PR #100, opened 2015-03-30
SOL-596 added the date range for querying the notifications

chrisndodge:


chrisndodge:
@muhhshoaib FYI here is the new PR to merge against a new long-running development branch.

edx_notifications/stores/sql/store_provider.py
```diff
@@ -184,6 +186,9 @@ def _get_prepaged_notifications(self, user_id, filters=None, options=None):
         if type_name:
             query = query.filter(msg__msg_type=type_name)
 
+        if start_date and end_date:
+            query = query.filter(created__gte=start_date, created__lt=end_date + timedelta(days=1))
```
chrisndodge: we should drop the timedelta. start_date and end_date should specify the exact range to query against.

edx_notifications/stores/sql/tests/test_store_provider.py
```diff
@@ -578,6 +578,43 @@ def test_get_notifications_for_user(self):
...
+                filters={
+                    'start_date': msg1.created.date(),
+                    'end_date': msg2.created.date()
+                }
+            )
+            self.assertEqual(len(notifications), 2)
+            self.assertEqual(notifications[0].msg, msg2)
+            self.assertEqual(notifications[1].msg, msg1)
+
+        self.assertEqual(
```
chrisndodge: Can we write a test where msg2's create date is updated and then we write a query so that we only get msg1 back?

muhhshoaib: Ok

muhhshoaib:
@chrisndodge i have update the code and the test cases. Kindly review

edx_notifications/stores/sql/store_provider.py
```diff
@@ -184,6 +185,9 @@ def _get_prepaged_notifications(self, user_id, filters=None, options=None):
         if type_name:
             query = query.filter(msg__msg_type=type_name)
 
+        if start_date and end_date:
+            query = query.filter(created__gte=start_date, created__lte=end_date)
+
```
chrisndodge: Sorry - hate to say this, but we should be able to filter either by start_date *or* end_date *or* both. Something like:

if start_date:
   query = query.filter(created__gte=start_date)

if end_date:
   query=query.filter(created__lte=end_date)

And maybe add some additional tests to:

filter only by start_date
filter only by end_date

Thanks.

chrisndodge:
One thing to address, otherwise this looks right. Thanks.

muhhshoaib:
@chrisndodge i have updated the changes.

chrisndodge:
@muhhshoaib looks good. Thanks!

PR #104, opened 2015-03-31
Afzaledx/sol 597 digests timer callback handler

chrisndodge:


edx_notifications/digest_timers.py
```diff
@@ -0,0 +1,35 @@
...
+
+DAILY_DIGEST_TIMER_NAME = 'daily-digest-timer'
+WEEKLY_DIGEST_TIMER_NAME = 'weekly-digest-timer'
+
+
+@receiver(perform_timer_registrations)
+def register_digest_timers():
+    store = notification_store()
+
+    weekly_digest_timer = NotificationCallbackTimer(
```
chrisndodge: Can you first look to see if this timer name already exists? I think if we don't do this, every time the server restarts, it's going to reset the timer. There's a method that I think is called store.get_notification_timer(timer_name).

edx_notifications/startup.py
```diff
@@ -26,4 +26,8 @@ def initialize():
     # notification types
     perform_type_registrations.send(sender=None)
 
+    # alert the application tiers that they should register their
+    # notification timers/callbacks
+    perform_timer_registrations.send(sender=None)
```
chrisndodge: If you use this signaling approach, it's an unfortunate thing in python that you will also have to put up at the top:

import digest_timers  # pylint: disable=unsed-import   (I think that's the pylint error override, see a similar example)

Otherwise, python will not load that digest_timer.py module before perform_timer_registrations.send() is called and the receiver that is defined there will not get called.

afzaledx:
Looks like the failing build is somehow related to travis/astroid as mentioned here:

https://bitbucket.org/logilab/astroid/issue/84/135-crashes-with-typeerror-_yes-object-is

This was set to resolved 4 days ago. Do we have to update our travis environment?

chrisndodge:
@afzaledx I explicitly added (and pinned) astroid to 1.3.4 which is what @nedbat's team did. Seems to make the build run successfully now.

chrisndodge:
@afzaledx now that we got builds working again, is this ready for review or do you still have some work to do on it? Thanks.

edx_notifications/callbacks.py
```diff
@@ -185,3 +185,16 @@ def _send_to_scoped_users(msg, scope_name, scope_context):
     )
 
     return num_dispatched
+
+
+class NotificationDigestMessageCallback(NotificationCallbackTimerHandler):
+    """
+        This is called by the NotificationTimer when there is a
+        timed notification that needs to be dispatched
```
chrisndodge: Can you update the doc string to reflect the use case here

edx_notifications/digest_timers.py
```diff
@@ -0,0 +1,47 @@
...
+
+    try:
+        store.get_notification_timer(DAILY_DIGEST_TIMER_NAME)
+    except ItemNotFoundError:
+        daily_digest_timer = NotificationCallbackTimer(
+            name=DAILY_DIGEST_TIMER_NAME,
+            callback_at=datetime.datetime.now(pytz.UTC),
+            class_name='edx_notifications.callbacks.NotificationDigestMessageCallback',
+            is_active=True,
+            periodicity_min=24 * 60
```
chrisndodge: Maybe for better readability 24 * 60 should be a global const like DAILY_IN_MINS

edx_notifications/digest_timers.py
```diff
@@ -0,0 +1,47 @@
...
+
+    try:
+        store.get_notification_timer(WEEKLY_DIGEST_TIMER_NAME)
+    except ItemNotFoundError:
+        weekly_digest_timer = NotificationCallbackTimer(
+            name=WEEKLY_DIGEST_TIMER_NAME,
+            callback_at=datetime.datetime.now(pytz.UTC),
+            class_name='edx_notifications.callbacks.NotificationDigestMessageCallback',
+            is_active=True,
+            periodicity_min=7 * 24 * 60
```
chrisndodge: Likewise, Maybe for better readability 7 * 24 * 60 should be a global const like WEEKLY_IN_MINS

edx_notifications/base_data.py
```diff
@@ -210,7 +210,7 @@ class DateTimeField(TypedField):
     Specialized subclass of TypedField(datetime) as a convienence
     """
 
-    _expected_types = [datetime]
+    _expected_types = [datetime, FakeDatetime]
```
chrisndodge: Hmmm. This is an unfortunate artifact of mocking out the system timer. Not sure if there's something better to do here....

chrisndodge:
@afzaledx just a few small items. Otherwise this looks good.

edx_notifications/base_data.py
afzaledx: This seemed to be the smallest change to get around the issue. It rejected the object otherwise (in `__set__`, it raised a `TypeError` `TypeError: Field expected type of '[<type 'datetime.datetime'>]' got '<class 'freezegun.api.FakeDatetime'>'`). I guess no big harm in letting it stay.

chrisndodge:
Looks good. Merging to the main feature branch.

edx_notifications/digest_timers.py
```diff
@@ -0,0 +1,49 @@
...
+    receiving the 'perform_timer_registrations' signal)
+    """
+    store = notification_store()
+
+    try:
+        store.get_notification_timer(DAILY_DIGEST_TIMER_NAME)
+    except ItemNotFoundError:
+        daily_digest_timer = NotificationCallbackTimer(
+            name=DAILY_DIGEST_TIMER_NAME,
+            callback_at=datetime.datetime.now(pytz.UTC),
```
chrisndodge: Even though I just merged this in, I'm wondering if we need to set the callback_at (here and below) to be 00:00 UTC the following day. The way that it is now, the digest will be produced at the same time every day that the service first started up....

Can you change?

PR #106, opened 2015-03-31
Revert "bugs fixes as highlighted by Kashif in edx-notifications"

chrisndodge:
Reverts edx/edx-notifications#98


chrisndodge:
@asadiqbal08 sorry, I had to revert your PR. I believe (but I'm not sure) that after merging it in all builds on cdodge/mcka-master started failing. I think there might be an edge case or other infinite loop that can happen in your change.

PR #116, opened 2015-04-02
Set the first execution at the upcoming midnight UTC.

afzaledx:


afzaledx:
Stalled build.

chrisndodge:
Looks like you need to update unit tests, also

PR #117, opened 2015-04-02
SOL-600 added HTTP endpoints for User Preferences

muhhshoaib:
@chrisndodge Kindly review the changes.

- added the new endpoints for UserPreferences Tables.
- added the test cases for all the endpoints.

### DIFF
--- a/edx_notifications/server/api/consumer.py
+++ b/edx_notifications/server/api/consumer.py
@@ -8,14 +8,17 @@ from rest_framework import status
 from rest_framework.response import Response
 
 from django.http import Http404
+from edx_notifications.data import UserNotificationPreferences
 
 from edx_notifications.lib.consumer import (
     get_notifications_count_for_user,
     get_notifications_for_user,
+    get_notification_preferences,
     get_notification_for_user,
     mark_notification_read,
-    mark_all_user_notification_as_read
-)
+    mark_all_user_notification_as_read,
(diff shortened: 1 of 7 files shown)


edx_notifications/server/api/consumer.py
```diff
@@ -238,6 +240,79 @@ def post(self, request):
...
+        try:
+            # this will raise an ItemNotFoundError if the user_id/name combo
+            # cannot be found
+            user_preference = get_user_preferences_with_name(int(request.user.id), name)
+        except ItemNotFoundError:
+            raise Http404()
+
+        return Response([user_preference.get_fields()], status.HTTP_200_OK)
+
+    def post(self, request, name):
```
chrisndodge: Hi @muhhshoaib . Sorry, perhaps I did not describe this correctly in the JIRA ticket. This POST is for a user to set his/her preference. So POST parameters should only be 'value' (we don't need display_name, display_description, and default value). In this handler all we should need to do is call set_user_preference(), as this API endpoint can only be called for a user to set an already existing Preference (not create a new one).

For test cases, we should call this endpoint with both "create" (first time user sets preference) as well as "update" (changing an existing UserPreference) cases.

Hope that makes sense, sorry for any confusion.

muhhshoaib:
@chrisndodge can you take a quick look into the post functionality and let me know if anything is missing 

Thanks

PR #118, opened 2015-04-02
Added system-defined Notification Preferences.

afzaledx:


chrisndodge:
Builds are failing because of pylint errors.

chrisndodge:
Looks right otherwise.

PR #119, opened 2015-04-03
Notifications - Auto refresh behaviour on view all tab

asadiqbal08:
JIRA Story : SOL-640. 
Description: If a notification is generated when the view all tab is open, then notification would be visible in the view all section as well. 

@chrisndodge kindly look into that.


chrisndodge:
Hi @asadiqbal08 I'm going to merge this into the "digests" working branch. We're releasing Notifications to production, and I don't think this is a severe enough fix to merge it into master. FYI @kashifch

PR #126, opened 2015-04-07
SOL-660 Archive deleted UserNotifications

asadiqbal08:
@chrisndodge kindly look into that.


edx_notifications/stores/sql/models.py
```diff
@@ -402,3 +428,13 @@ def load_from_data_object(self, notification_timer):
...
         self.err_msg = notification_timer.err_msg
         self.results = DictField.to_json(notification_timer.results)
+
+
+@receiver(pre_delete, sender=SQLUserNotification)
+def archive_deleted_user_notification(sender, instance, *args, **kwargs):  # pylint: disable=unused-argument
+    """
+    Archiving the deleted user notifications.
+    """
+    notification_archive_obj = SQLUserNotificationArchive()
```
chrisndodge: Can you add a configuration flag here to check if the installation wishes to actually archive notifications. I say this because many installations of Open edX might not wishes to keep all these in a database, since the database table will grow very large over time.

Can you put this configuration in const.py and have it default to "False" (don't archive messages by default)? Thanks.

chrisndodge:
Thanks looks good overall. Just add a configuration flag to say if we want to archive notifications or not.

Also, can you add a unit test to assert that UserNotification and UserNotificationArchive have the same schema (be sure to look in both directions). One concern that I have is that - in the future - we'll add an additional field to UserNotification but forget to add the same field in UserNotificationArchive. Having a unit test to compare will automatically catch this issue.

chrisndodge:
Sorry, seems like a merge conflict now. It's probably the change in const.py. Should be simple to rebase and resolve.

chrisndodge:
@asadiqbal08 I just did the rebase, I'll merge if the build passes

PR #128, opened 2015-04-07
force purge Django Management command.

afzaledx:
@chrisndodge, added the django management command to force-purge the notifications. : SOL-661

### DIFF
--- a/edx_notifications/management/commands/tests/test_background_check.py
+++ b/edx_notifications/management/commands/tests/test_background_check.py
@@ -9,7 +9,7 @@ from datetime import datetime, timedelta
 from django.test import TestCase
 from django.dispatch import receiver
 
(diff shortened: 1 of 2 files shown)


edx_notifications/management/commands/tests/test_background_check.py
```diff
@@ -67,3 +67,16 @@ def test_timer_execution(self):
...
+    """
+    Test suite for the management command
+    """
+
+    def test_purge_command_check(self):
+        """
+        Invoke the Management Command
+        """
+
+        force_purge.Command().handle()
```
chrisndodge: hate to say this, but this isn't a real test. It certainly will get us some code coverage stats. :-) 

Can you set up some test data and make sure it got deleted? Thanks.

PR #138, opened 2015-04-14
(WIP) SOL-602- added the groupings for user notifications

muhhshoaib:
@chrisndodge added the groupings for the user notifications.

Since underscore.py is not working on the if conditions of the underscore templates as @afzaledx sent in the email. I have removed the if conditions from the announcement and forums underscore templates so that i can work on the groupings and sorting stuff as we did in the backbone unread view. 

This is a work in progress.

Thanks


chrisndodge:
Thanks. I'm thinking we should keep that 'if' statement removed from the Underscore templates and then - if UI want to hide them - we can use a CSS rule to 'display: none'. I'm going to investigate this more.

chrisndodge:
@muhhshoaib hey there, I did some touch up work on this branch, so please fetch latest:

- per wishes of McKA, if there are no unread notifications during the digest period, don't send an email at all. This is a configuration setting to change behavior, if one wishes.

- moved some configuration over to const.py so we can override in some external settings file.

- pass along a 'subject' and 'from_email' which will be used in the sending of the email

- update .css rule (in testserver) to target the 'date' field which should be hidden when on the 'all' pane in backbone (we have to copy this over to McKA Apros - I wrote up a task and assigned to me).

edx_notifications/renderers/basic.py
```diff
@@ -74,11 +77,17 @@ def render(self, msg, render_format, lang):
                 self.underscore_template = us.template(template_string)
 
         _payload = copy.deepcopy(msg.payload)
+
+        if datetime.datetime.date(msg.created) == datetime.datetime.now(pytz.UTC).date():
+            created_str = 'Today at ' + msg.created.strftime("%H:%M%p")
+        else:
+            created_str = msg.created.strftime("%B %d, %Y") + ' at ' + msg.created.strftime("%H:%M%p")
```
chrisndodge: I think we shouldn't translate this to today mainly because all datetime's are in GMT and this could look a little funny to users. I wrote up a story whether we should translate digest timestamps to localtime (based on a browser telling us what timezone the user last logged in at). But this needs to be priorities.

So, in the meantime, remove the 'if' statement and not use the 'Today' notation.

edx_notifications/server/web/templates/django/digests/unread_notifications_inner.html
```diff
@@ -0,0 +1,20 @@
...
+                    <li class="xns-item">
+                        <div class="xns-item-body">
+                            <span class='xns-{{ message.group_name }}'>{% autoescape off %}{{ message.html }}{% endautoescape %}</span>
+                        </div>
+                    </li>
+                {% endfor %}
+            {% endfor %}
+        {% endif %}
+    </ul>
+</div>
```
chrisndodge: FYI, I renamed this file (and moved it as well)

testserver/settings.py
```diff
@@ -70,6 +70,8 @@
     'django.template.loaders.app_directories.Loader',
 )
 
+EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
```
chrisndodge: Do you know how we can actually tie into some SMTP server? While it is OK to look at the console, I'd like us to manually test actually receiving the email in various clients. Maybe we should look to see if we could use our personal gmail accounts as a SMTP server?!?

chrisndodge:
Just a few things. Not sure what we should do about code coverage. Do you think you can get some unit test coverage quickly in here?

testserver/settings.py
chrisndodge: see: http://stackoverflow.com/questions/6914687/django-sending-email

chrisndodge:
Actually, I'm not thinking we should get to feature complete and then add unit tests.

chrisndodge:
@afzaledx thanks for getting coverage up. There might be some touchup work later, but merging this for now.

PR #142, opened 2015-04-16
Afzaledx/sol 721 digest mgmt command

afzaledx:


afzaledx:
@chrisndodge, I have the structure of the management command in place. It takes in --daily, --weekly and --ns=NAMESPACE (default is ALL if no namespace specified). Somehow, the actual emails are not going through at this point though. Will look into this with more detail tomorrow, and will try to write tests as well.

chrisndodge:
Thanks. I'll take a look. When you say "not going through" is it just going through the SMTP server? If you set your emailer to print to the console, does it work? I'll give it a try later on today.

edx_notifications/const.py
```diff
@@ -54,7 +54,7 @@
     "NOTIFICATION_DAILY_DIGEST_SUBJECT",
     "Please set NOTIFICATION_DAILY_DIGEST_SUBJECT in Django settings"
 )
-NOTIFICATION_DAILY_DIGEST_SUBJECT = getattr(
+NOTIFICATION_WEEKLY_DIGEST_SUBJECT = getattr(
```
chrisndodge: oops. Thanks for catching that

afzaledx:
The email goes through when I click from the test server, but I am using the same --settings=testerver.settings but the email doesn't show up. Haven't had the chance to debug properly. I will fix that, but for now I just want you comments on the structure of the command i.e. arguments etc.

edx_notifications/management/commands/tests/test_background_check.py
```diff
@@ -150,3 +150,11 @@ def test_purge_command_check(self):
             ),
             0
         )
+
+
+class ForceSendDigestCommandTest(TestCase):
```
chrisndodge: Can you put this TestCase in its own test file?

chrisndodge:
@afzaledx just a heads up that I've put some commits on this branch. Basically, I wanted to be able for us to send both digests for unread as well as all notifications. I needed to do the work here, because otherwise it would have conflicted. So you will need to pull latest before doing any additional work.

chrisndodge:
Closing this PR because I've opened up a new PR which pulls out the force_send_digest command (which has been moved to edx-platform): https://github.com/edx/edx-notifications/pull/145

PR #144, opened 2015-04-17
Cdodge/support all notification digests

chrisndodge:


chrisndodge:
@afzaledx FYI, I've cherry-picked the commits from your django management command to allow for sending digests of all notifications (not just unread). I'll move your management command over to a PR on edx-platform since we need to define that at the app tier where the various resolvers live.

PR #149, opened 2015-04-22
SOL-775 Notifications - "Mark as Read" tab becomes unresponsive

afzaledx:
Catered for a condition where the call to "mark all read" was being skipped since the collection had changed. Modified the check to enable the call to go through while still blocking multiple clicks on "Mark unread" from being passed to the server.


afzaledx:
@chrisndodge, please have a look.

PR #150, opened 2015-04-24
(WIP) SOL-747 and SOL-748

muhhshoaib:
@chrisndodge Here is the PR for the triggeredEmail channel work. 

- we have renamed the durable channel in settings to triggeremail channel for testing purposes. 
- There is not support for multiple channels right now so we changed the channel name for testing purposes.


chrisndodge:
@muhhshoaib @afzaledx , I reworked the basics and tied into the test server so one can publish to the triggered email channel. Please review the changes to learn them and continue on the story which will be to actually generate the HTML email and send it.

muhhshoaib:
@chrisndodge I have added the outer template for the triggered email and added the test case to make sure that coverage won't decrease. The test case is simple, i am just checking the count on the bulk dispatch notification using the triggered_email channel. Let me know if anything is missing.

chrisndodge:
@muhhshoaib tried it out and it looks good. There will probably be some touchup work on styling and content, but I think this is good enough to merge into the main working digest branch.

chrisndodge:
One last comment, we should probably figure out a way to patch/mock the sending of the email to be able to write some tests to assert that the content that is being sent is what we expect.

PR #154, opened 2015-04-28
(WIP) SOL-736- added the abstract methods for nosql store provider

muhhshoaib:
@chrisndodge here are the changes.


testserver/settings.py
```diff
@@ -120,8 +120,11 @@
 NOTIFICATION_PREFERENCES_IS_VISIBLE = True
 
 NOTIFICATION_STORE_PROVIDER = {
-    "class": "edx_notifications.stores.sql.store_provider.SQLNotificationStoreProvider",
+    "class": "edx_notifications.stores.mongo.store_provider.MongoNotificationStoreProvider",
     "options": {
+        'host': 'localhost',
+        'port': 27017,
+        'database_name': 'notifications'
     }
```
chrisndodge: Hmmm. Let me think if there's a way we can have configuration definitions for different providers at the same time. I might make a commit here.

edx_notifications/stores/mongo/store_provider.py
```diff
@@ -0,0 +1,140 @@
...
+class MongoNotificationStoreProvider(SQLNotificationStoreProvider):
+    """
+
+    """
+
+    def __init__(self, **kwargs):
+        super(MongoNotificationStoreProvider, self).__init__(**kwargs)
+        self.client = MongoClient(kwargs.get('host'), kwargs.get('port'))
+        self.db = self.client[kwargs.get('database_name')]
+        self.collection = self.db.user_notification
```
chrisndodge: I believe here we'll have to make sure that indexes exist in the collection

edx_notifications/stores/mongo/store_provider.py
```diff
@@ -0,0 +1,140 @@
...
+                    'user_notification':
+                        {
+                            'msg_id': user_msg.msg.id,
+                            'created': datetime.datetime.now(pytz.UTC),
+                            'modified': datetime.datetime.now(pytz.UTC),
+                            'user_context': user_msg.user_context if user_msg.user_context else None,
+                            'read_at': user_msg.read_at if user_msg.read_at else None
+                        }
+                }}
+            )
```
chrisndodge: I think we can use "upsert" semantics here so we don't need to do this existence test (which - since we are not locking databases - can lead to race conditions). See: http://api.mongodb.org/python/current/api/pymongo/collection.html

edx_notifications/stores/mongo/store_provider.py
```diff
@@ -0,0 +1,140 @@
...
+            user_notification = self.collection.insert(
+                {
+                    'user_id': user_msg.user_id,
+                    'user_notification': [{
+                        'msg_id': user_msg.msg.id,
+                        'created': user_msg.created if user_msg.created else datetime.datetime.now(pytz.UTC),
+                        'modified': datetime.datetime.now(pytz.UTC),
+                        'user_context': user_msg.user_context if user_msg.user_context else None,
+                        'read_at': user_msg.read_at if user_msg.read_at else None
+                    }]
```
chrisndodge: I think we'll need to duplicate data and put the msg_type in here as well in order to do all of the filtering we need to do in _get_prepaged_notifications. Luckily, once set, msg_type does not change.

chrisndodge:
@muhhshoaib @afzaledx for tests, I believe you should be able to simply copy what is in sql/tests/test_store_provider.py and make some modifications. 

You'll need to remove all of the self.assertNumQueries() - or (better yet) at least update them with the new Django ORM query counts.

edx_notifications/stores/mongo/store_provider.py
```diff
@@ -0,0 +1,141 @@
...
+        if type_name:
+            query_object.update({'user_notification.msg_type_name': type_name})
+
+        if start_date:
+            query_object.update({'created': {"gte": start_date}})
+
+        if end_date:
+            query_object.update({'created': {"lte": end_date}})
+
+        collection = self.collection.find(query_object)
```
chrisndodge: Sorry, just realized: we need to sort by create date DESC here. Also, the interface is supposed to support OFFSET and LIMIT semantics as well. Not sure how well that maps to MongoDB API, but it might have to be done post-processing.

Either way, we'll need to have the first record in this collection be the most recent one. I'll also add a note below about this.

edx_notifications/stores/mongo/store_provider.py
```diff
@@ -0,0 +1,141 @@
...
+        Create or Update the mapping of a user to a notification.
+        """
+
+        user_notification = self.collection.update(
+            {
+                'user_id': user_msg.user_id,
+            },
+            {
+                '$set': {'user_id': user_msg.user_id},
+                '$push': {
```
chrisndodge: In terms of default ordering where we want the most recent notification first, can we insert the new user_notification at the beginning of the sub-collection array? Sorry, I'm not sure if $push here means 'append at the end' or 'insert at the beginning'.

chrisndodge:
I just thought of an index we'll need: we need an index on the msg_id and readat time stamp inside every user msg. At some point in the future we will need to query how many unread instances of a particular msg_id there is so that we can add some clean up tasks to remove msgs that no longer have unread pointers to them.

chrisndodge:
Hrm. Sorry, I didn't notice this PR is against notification-digest, my intent was for this to be against master. I'll open a new PR.

PR #162, opened 2015-09-09
Registering GP v2 notification types

e-kolpakov:
**Background:** Group Project XBlock sends notifications for file upload, grades posted and stage open/close dates. Group Project v2 sends the same types of messages, but uses different navigation approach, so the URLs in messages need to be different. For this to work, GP v2 has to send messages of unique types. 

**Description:** This PR registers new types of messages with notifications service.

**Dependencies:** This PR is one of the three that fix notifications in Group Project v2 XBlock; the other two are:
* https://github.com/edx-solutions/edx-platform/pull/517 - adds URLResolvers
* https://github.com/open-craft/xblock-group-project-v2/pull/53 - switches GP v2 to use dedicated message types

**Related:** [YONK-146](https://openedx.atlassian.net/browse/YONK-146) was caused by changing URL resolvers for GP v1. Should not be affected by this change.

**Testing instructions:** see testing instructions on https://github.com/edx-solutions/edx-platform/pull/517

### DIFF
--- a/edx_notifications/openedx/group_project.py
+++ b/edx_notifications/openedx/group_project.py
@@ -12,6 +12,21 @@ from edx_notifications.renderers.basic import UnderscoreStaticFileRenderer
 
 from django.dispatch import receiver
 
+GROUP_PROJECT_V1_NOTIFICATION_PREFIX = u'open-edx.xblock.group-project'
+GROUP_PROJECT_V2_NOTIFICATION_PREFIX = u'open-edx.xblock.group-project-v2'
+
+GROUP_PROJECT_RENDERER_PREFIX = 'edx_notifications.openedx.group_project'
+
+
+class NotificationMessageTypes(object):
+    """
+    Message type constants
+    """
+    STAGE_OPEN = u'stage-open'
+    STAGE_DUE = u'stage-due'
+    FILE_UPLOADED = u'file-uploaded'
+    GRADES_POSTED = u'grades-posted'
+
 
 class GroupProjectFileUploadedRenderer(UnderscoreStaticFileRenderer):
     """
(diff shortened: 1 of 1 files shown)


e-kolpakov:
@chrisndodge not totally sure what's the process of reviewing contributions in this PR is. Should I create a ticket in JIRA somewhere? If so, which project is used? And I'm pinging you because you've got the most commits in the repository, so if you're not the right person to discuss that kind of things - please point me out to who should I talk to.

Kelketek:
@e-kolpakov You have a missing docstring which is causing a quality build failure here.

edx-webhook:
Thanks for the pull request, @e-kolpakov! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=162&repo=edx%2Fedx-notifications

edx_notifications/openedx/group_project.py
```diff
@@ -48,31 +63,25 @@ def register_notification_types(sender, **kwargs):  # pylint: disable=unused-arg
...
+                renderer="{prefix}.{renderer}".format(prefix=GROUP_PROJECT_RENDERER_PREFIX, renderer=renderer),
+            )
         )
-    )
 
-    register_notification_type(
-        NotificationType(
-            name=u'open-edx.xblock.group-project.grades-posted',
-            renderer='edx_notifications.openedx.group_project.GroupProjectGradesPostedRenderer',
+        # GP v2 can reuse GP v1 renderers
```
chrisndodge: Just double confirming that there are no schema changes in the notification payloads.

However, if they are identical to GP v1 notifications, why do you want to register different notification types?

e-kolpakov: @chrisndodge there are no schema changes in payloads.

The thing is, the notifications should lead to a slightly different url: namely, GPv2 uses `activate_block_id` feature to allow navigating to various types of nested blocks. This can't be resolved client-side, because `activate_block_id` is passed all the way down to the XBlock being rendered.

The reason why I had to do this is that we still need to support GP v1 with old-style links, and GP v2 with new links. This could be done via URL resolvers, but (if I do understand right) URl resolver is applied to all messages of some type. In order to apply different resolver one needs a dedicated message type. So here they are :)

chrisndodge: Thanks, yep, I understand now. URL resolvers are bound to a MessageType.

Kelketek:
:+1:

chrisndodge:
I think this should be fine, so +1. Just so you know one doesn't have to put the renderer registrations inside edx-notifications repo - although I'm not sure we've tried elsewhere. We chose to put all notification type registrations here so that the test Django server (which is part of this repo) would have those definitions and we could write bokchoy tests against it.

e-kolpakov:
@chrisndodge I can't say I've made an exhaustive search, but I haven't seen any renderer registrations anywhere else. Also, is there a way to register notifications outside of edx-notifications? And in general, is there developer documentation on edx-notifications somewhere already? Google only finds a page that promise dev documentation to be released Q1 2016.

PR #166, opened 2017-02-21
YONK-521: fixed duplicate preferences issue

ziafazal:
This PR has changes to fix a issue related to duplicate user preference data causing breakage of weekly and daily notification digest workflow. This PR 

1. adds a unique constraint for user and preference fields in `SQLUserNotificationPreferences` model. 

2. Adds a management command to clear duplicates in DB for `SQLUserNotificationPreferences` model. 

3. Adds ability to filter namespaces by start and end datetimes which would optimize time to send digests

4. Fixes a broken unit test `edx_notifications.tests.test_digests.DigestTestCases#test_default_group_mapping`

@ayesha-baig would you please review it?


ayesha-baig:
@ziafazal, the script does not handle the case where there are more than two records with same user and preference.

ziafazal:
@ayesha-baig added support to remove more than one duplicate preferences.

PR #167, opened 2017-02-27
added command to remove duplicate user prefs data

ziafazal:
@fredsmith this is the command we are planning to add to cypress release of McKA. We need to update the hash of notifications app in edx-platform after merging it. I would recommend testing it on QA before moving it to Prod since we tested this command on Euclayptus which has Django 1.8 whereas Cypress has django 1.4


ziafazal:
closing this in favour of PR #168

PR #170, opened 2017-08-03
Push notification through Urban Airship API

jawadtahir:


edx_notifications/channels/channel.py
```diff
@@ -235,6 +235,16 @@ def bulk_dispatch_notification(self, user_ids, msg, exclude_user_ids=None, chann
         raise NotImplementedError()
 
     @abc.abstractmethod
+    def dispatch_notification_to_tag(
```
ziafazal: I think we don't need to define a new method in `BaseNotificationChannelProvider` we should use existing `bulk_dispatch_notification` method to send notifications to a tag group if it is available in `channel_context`.

edx_notifications/channels/tests/test_urban_airship.py
```diff
@@ -0,0 +1,48 @@
...
+            renderer='edx_notifications.openedx.course_announcements.NewCourseAnnouncementRenderer'
+        )
+        register_notification_type(self.msg_type)
+        self.msg = NotificationMessage(
+            namespace='foo/bar/baz',
+            msg_type=self.msg_type,
+            payload=CANNED_TEST_PAYLOAD['open-edx.studio.announcements.new-announcement']
+        )
+
+    def test_publish_notification_tag(self):
```
ziafazal: we should add tests for other user cases like sending push notification to single user etc.

edx_notifications/channels/urban_airship.py
```diff
@@ -0,0 +1,131 @@
...
+import json
+import logging
+
+import requests
+from requests.auth import HTTPBasicAuth
+
+from edx_notifications.channels.channel import BaseNotificationChannelProvider
+
+# system defined constants that only we should know about
+UA_API_ENDPOINT = 'https://go.urbanairship.com/'
```
ziafazal: we should change it to
```
UA_PUSH_API_ENDPOINT = 'https://go.urbanairship.com/api/push'
```

edx_notifications/channels/urban_airship.py
```diff
@@ -0,0 +1,131 @@
...
+HEADER_DICT = {
```
ziafazal: nit: Can we rename it to `HTTP_HEADERS`?

edx_notifications/channels/urban_airship.py
```diff
@@ -0,0 +1,131 @@
...
+        :param tag:
+        :param msg:
+        :return:
+        """
+        # Tag validation
+        if group is None or len(group) < 1:
+            raise Exception('No tag is provided')
+
+        obj = {
+            'notification': {'alert': msg.payload['excerpt']},
```
ziafazal: Can assert present of `excerpt` in payload before using it?

edx_notifications/channels/urban_airship.py
```diff
@@ -0,0 +1,131 @@
...
+            'device_types': 'all',
+            'audience': {'group': group}
+        }
+        if tag:
+            (obj['audience'])['tag'] = tag
```
ziafazal: nit: what is the reason for using `()` around `obj` dict?

edx_notifications/channels/urban_airship.py
```diff
@@ -0,0 +1,131 @@
...
+
+        # Send request to UA API
+        resp = requests.post(
+            UA_API_ENDPOINT + 'api/push',
+            data=obj,
+            headers=HEADER_DICT,
+            auth=HTTPBasicAuth(self.application_id, self.rest_api_key)
+        )
+
+        resp = resp.json()
```
ziafazal: Can we do some error handling here and if there is error add it to error logs.
We first need to valid HTTP response codes and if we have 200 HTTP response code we can further evaluate the response to see value of `ok` key for any errors.
https://docs.urbanairship.com/api/ua/#error-object

settings.py
```diff
@@ -141,6 +148,7 @@
 # list all of the mappings of notification types to channel
 NOTIFICATION_CHANNEL_PROVIDER_TYPE_MAPS = {
     '*': 'durable',  # default global mapping
+    'open-edx.studio.announcements.new-announcement': 'urban-airship'
```
ziafazal: This should in edx-platform too.

settings.py
```diff
@@ -152,3 +160,4 @@
 
 # digest email css
 NOTIFICATION_DIGEST_EMAIL_CSS = 'edx_notifications/css/email_digests.css'
+MCKA_APP_URL = 'https://www.mckinseyacademy.com/'
```
ziafazal: We probably should have a better way to determine the domain of the environment notifications app is running in.

edx_notifications/channels/urban_airship.py
```diff
@@ -0,0 +1,149 @@
...
+        :param user_id:
+        :param msg:
+        :param channel_context:
+        :return:
+        """
+
+        obj = {
+            'notification': {'alert': msg.payload['excerpt']},
+            'audience': {'named_user': user_id},
+            'device_types': ['ios', 'android']
```
jawadtahir: You will also need to explicitly set the device_types because Named Users are not supported on every platform
Ref: https://support.urbanairship.com/hc/en-us/community/posts/208269303-Target-Named-User

ziafazal:
@jawadtahir closing this in favour of #171

PR #173, opened 2017-10-30
add all users payload and pass auth_dict as a param in order to get a…

ihtram:
…pp_id and secret key

### DIFF
--- a/edx_notifications/channels/urban_airship.py
+++ b/edx_notifications/channels/urban_airship.py
@@ -57,7 +57,8 @@ class UrbanAirshipNotificationChannelProvider(BaseNotificationChannelProvider):
         """
         payload = self.create_payload(msg, user_id)
         payload = json.dumps(payload)
-        return self.call_ua_push_api(payload)
+        api_credentials = channel_context.get('api_credentials') if channel_context else None
+        return self.call_ua_push_api(payload, api_credentials)
 
     @staticmethod
     def create_payload(msg, user_id):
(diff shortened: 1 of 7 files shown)


edx_notifications/channels/urban_airship.py
```diff
@@ -102,18 +104,21 @@ def call_ua_push_api(self, payload):
...
         :param user_ids:
         :param msg:
         :param exclude_user_ids:
         :param channel_context:
+        :param send_to_all:
         :return:
         """
         if 'tag_group' in msg.payload:
             payload = self.create_tag_group_payload(msg)
+        elif send_to_all:
```
ziafazal: `send_to_all ` should be a flag in `msg.payload` not a parameter because we should not change the signature of `bulk_publish_notification_to_users ` method.

edx_notifications/channels/urban_airship.py
```diff
@@ -77,19 +77,21 @@ def create_payload(msg, user_id):
 
         return obj
 
-    def call_ua_push_api(self, payload):
+    def call_ua_push_api(self, payload, auth_dict):
```
ziafazal: `auth_dict` should be renamed to `api_keys` should be a pass in via `channel_context` parameter because we should not change the signature of `bulk_publish_notification_to_users ` method.

ihtram:
@ziafazal i have implemented all the mentioned changes. can you please review again?

PR #174, opened 2017-11-01
Ihtram/mobileapps notification receiver

ihtram:


ihtram:
@ziafazal  can you please review it?

PR #177, opened 2017-12-20
file name fix and pylint issues

Ayub-Khan:
File name fix.


nedbat:
@Ayub-Khan Hmm, this turned into a bigger change, I guess.  I'm a little surprised to see name changes, and pylint pragmas about name changes on the same lines.  Next time, let's talk about the best way to deal with those issues.

PR #191, opened 2019-03-21
fix visit notification non working link on Discussion page

umarmughal824:
MCKIN-9970

On Notification page, Notification were not navigating to their associated links. It's a fix for this issue.

When we try to visit the notification, It send AJAX request to server to mark that specific notifictaion **read** while processing this request, it does not pass the **CSRF Token** in the request on **Discussion page**. This is a fix for it.

PR #192, opened 2019-04-12
MCKIN-10378 - CSRF fix for mark as read

naeem91:
This is the same fix as applied in #191 but was missed for other instances.

CSRF does not get passed when using `headers` now it is passed by beforeSend method of JQuery AJAX


naeem91:
@ihtram Could you please review?

PR #197, opened 2019-08-28
MCKIN-11422 Upgrade pynliner to 0.8.0

mudassir-hafeez:
For upgrading python 2.7 to 3.5 on Apros, pynliner needs to be upgraded to 0.8.0 as it's compatible with python 3.5.


edx-webhook:
Although this pull request is already closed, I've created [OSPR-5420](https://openedx.atlassian.net/browse/OSPR-5420) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #198, opened 2019-08-29
MCKIN-11421 Added support for python 3.5

nasirhjafri:

PR #199, opened 2019-08-30
MCKIN-11429 Configure tox with edx-notifications

moeez96:
Ticket: https://edx-wiki.atlassian.net/browse/MCKIN-11429

Configure Tox with edx-notifications to support test suite on both python 2.7+Django1.11 and python 3.5+Django (to be added later once edx-notifications is updated to that python version.)


edx-webhook:
Thanks for the pull request, @moeez96! I've created [OSPR-3815](https://openedx.atlassian.net/browse/OSPR-3815) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

Feel free to add as much of the following information to the ticket:
- supporting documentation
- edx-code email threads
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).



We can't start reviewing your pull request until you've submitted a     [signed contributor agreement](https://open.edx.org/wp-content/uploads/2019/01/individual-contributor-agreement.pdf)     or indicated your institutional affiliation. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

natabene:
@ihtram Please review soon.

tox.ini
```diff
@@ -0,0 +1,15 @@
+[tox]
+envlist = py27-django111
```
ihtram: @moeez96 We need to configure tox for both py27 and py35. Please configure tox for py35 as well.

ihtram: also add allow_failures in travis config for py35 to ignore py35 failures for now.

.travis.yml
```diff
@@ -7,10 +7,12 @@ addons:
 
 env:
   - DJANGO_SETTINGS_MODULE=settings
+sudo: required
```
ihtram: @moeez96 Why do we need root privilege here?

tox.ini
moeez96: @ihtram py35 is not a part of travis env hence adding allow_failures in travis for a tox config doesnt seems well. How about we keep py27 for this pr in tox and add py35 in tox with the next PR which actually contains all py35 tests. What do you think?

.travis.yml
moeez96: Removed

tox.ini
ihtram: we should configure py35 for travis. Adding tox for only py27 is meaningless as the purpose of tox is to run multiple environments. After configuring tox for both versions we can then start fixing broken tests.

moeez96:
@ihtram Please review

ihtram:
@moeez96 We still need to configure py35 with travis. So basically this should run tests on both py27 and py35. py35 tests will fail but you can add configurations to allow_failures for py35. Once this is done you need to rebase the other PR(https://github.com/edx/edx-notifications/pull/198) where unit tests have been fixed but those tests are running only for py27. 
So basically this should add support to run tests with both versions of python something like this:
https://github.com/edx/ecommerce/blob/master/.travis.yml

moeez96:
@ihtram py35 addedd to travis, Rebasing to PR #198 now.

moeez96:
@ihtram PR rebased, Failing tests will be solved in separate PR
Cc: @nasirhjafri

moeez96:
@ihtram please have a look

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5418](https://openedx.atlassian.net/browse/OSPR-5418) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks again for your contribution.

PR #201, opened 2019-09-12
Do not allow Python 3.5 test failures

moeez96:
Remove allow_failures for python 3.5 from travis configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,6 @@ matrix:
         - tox -e py35
         - ./node_modules/gulp/bin/gulp.js test
         - bash ./run_bokchoy_tests.sh
-  allow_failures:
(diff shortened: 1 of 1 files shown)


moeez96:
@ihtram Please review

msaqib52:
@ihtram could you create a release tag for python 3 changes, so we can use it.

ihtram:
@msaqib52  Created a new release tag 1.0 with name "Added python3 support"

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5416](https://openedx.atlassian.net/browse/OSPR-5416) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks again for your contribution.

PR #202, opened 2019-09-17
Remove django_nose from prod settings

jdmulloy:
It seems this isn't actually used in Prod and it's causing issues upgrading to python 3. Management commands try to import django_nose which doesn't exist and crashes.


msaqib52:
@ihtram could you also create a release tag so we can update it in Apros, Thanks.

ihtram:
@msaqib52 release 1.0.1 has been created

PR #250, opened 2021-12-30
Fix CI workflow to run tests

aht007:
JIRA ISSUE: [BOM-3119](https://openedx.atlassian.net/browse/BOM-3119)

## Description
While shifting CI to GitHub Actions there was a syntax issue in the workflow which caused the tests command to be ignored and hence tests were not running. This PR solves that issue and also splits the workflows according to their use case.


Makefile
```diff
@@ -45,7 +47,7 @@ test-js: ## run tests using npm
 	-./node_modules/gulp/bin/gulp.js test
 
 test-python: clean ## run tests using pytest and generate coverage report
-	$(TOX)pytest
+	$(TOX)pytest --ignore=testserver/test/acceptance
```
aht007, line 50: Ignoring acceptance tests here because they will be run separately with bokchoy

.github/workflows/js-tests.yml
```diff
@@ -0,0 +1,31 @@
+name: Node CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
```
mraarif, line 8: we can just remove the `branches` option

testserver/settings.py
```diff
@@ -85,7 +85,7 @@
         'rest_framework.throttling.UserRateThrottle',
     ),
     'DEFAULT_THROTTLE_RATES': {
-        'user': '10/sec',
+        'user': '100/sec',
```
mraarif, line 88: why did we update this?

aht007: Some tests were failing with pytest and @awais786 bhai suggested that we should increase the throttle rate... Those tests passed after this change

.github/workflows/js-tests.yml
aht007: Doing this causes duplicate checks on PRs

awais786 (approved):

UsamaSadiq (approved):

.github/workflows/bok-choy-tests.yml
```diff
@@ -0,0 +1,54 @@
...
+        sudo apt-get install xvfb gettext
+        wget https://github.com/mozilla/geckodriver/releases/download/v0.25.0/geckodriver-v0.25.0-linux64.tar.gz
+        mkdir geckodriver
+        tar -xzf geckodriver-v0.25.0-linux64.tar.gz -C geckodriver
+        export PATH=$PATH:$PWD/geckodriver
+        export BOKCHOY_HEADLESS=true
+
+    - name: Install dependencies
+      run: |
+        pip install -r requirements/base.txt
```
mraarif, line 40: do we've to install python dependencies to run bok-choy tests?

.github/workflows/ci.yml
```diff
@@ -30,29 +26,42 @@ jobs:
...
-
-    - name: Install npm dependencies
-      run: npm install
+    - name: Set up test environment
+      run: |
+        sudo apt-get update
+        sudo apt-get install xvfb gettext
+        wget https://github.com/mozilla/geckodriver/releases/download/v0.25.0/geckodriver-v0.25.0-linux64.tar.gz
+        mkdir geckodriver
+        tar -xzf geckodriver-v0.25.0-linux64.tar.gz -C geckodriver
```
mraarif, line 35: why are we installing these dependencies? we're not running `bok-choy` tests in this workflow

.github/workflows/ci.yml
```diff
@@ -30,29 +26,42 @@ jobs:
...
     - name: Install dependencies
       run: |
         pip install -r requirements/base.txt
         pip install -r requirements/testing.txt
 
-    - name: Run Python and Quality Tests
+    - name: Install Firefox 61.0
+      run: |
+        sudo apt-get purge firefox
+        wget "https://ftp.mozilla.org/pub/firefox/releases/46.0/linux-x86_64/en-US/firefox-46.0.tar.bz2"
```
mraarif, line 47: same here why do we need to install firefox to run `python` unit tests and `quality`

.gitignore
```diff
@@ -59,3 +59,5 @@ acceptance_tests/*.png
 node_modules/
 
 testserver/local_settings.py
+
+venv
```
mraarif, line 63: should be `venv/`

.github/workflows/bok-choy-tests.yml
aht007: Yes we need the dependencies to run browser tests. Otherwise it raises Module not found errors

.github/workflows/ci.yml
aht007: Thats a mistake. I will resolve that

mraarif (approved):

PR #254, opened 2022-03-29
feat: Added support for node 16

Jawayria:


.github/workflows/js-tests.yml
```diff
@@ -13,17 +13,20 @@ jobs:
...
     - name: Setup Node.js
       uses: actions/setup-node@v1
       with:
-        node-version: 12
+        node-version: ${{ matrix.node }}
 
     - name: Install npm dependencies
-      run: npm install
+      run: |
+        echo -e '[url "https://github.com/"]\n  insteadOf = "git://github.com/"' >> ~/.gitconfig
```
Jawayria, line 28: Dependency installation failed with error `[The unauthenticated git protocol on port 9418 is no longer supported](https://stackoverflow.com/questions/70663523/the-unauthenticated-git-protocol-on-port-9418-is-no-longer-supported)` so added this line as a fix. [ref](https://stackoverflow.com/questions/70663523/the-unauthenticated-git-protocol-on-port-9418-is-no-longer-supported)

aht007 (approved):

.github/workflows/js-tests.yml
Jawayria: Yes replacing with karma-jasmine-jquery-2 worked

